PR #3, opened 2020-08-06
Add event ports for illumos

papertigers:
This adds the use of the preferred evented I/O mechanism on illumos/solaris. This also means `polling` should work on a a real Solaris box (as it doesn't have eventfd or epoll) although I have not tested in such an environment. 

I ran the test suite and verified everything passed. I also ran a modified version of a toy [tcp-proxy](https://github.com/papertigers/proxy-rs) that uses async-std to verify bits were able to flow in both directions with the included patch.

Note I have only tested with the "solaris" version of rustc on illumos.  I will update this PR shortly with testing notes from a rust nightly build that actually uses the illumos rustc target.

Also worth noting I chose to reuse the pipe method from kqueue instead of eventfd even though illumos supports it, so that a real Solaris build is able to take advantage of this crate.

```
root - rustdev ~/src/polling (git:illumos) # cargo test
   Compiling polling v0.1.0 (/root/src/polling)
    Finished test [unoptimized + debuginfo] target(s) in 1.14s
     Running target/debug/deps/polling-fbc74159dd2c3d72

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests polling

running 10 tests
test src/lib.rs - Poller::interest (line 229) ... ok
test src/lib.rs - Poller::interest (line 240) ... ok
test src/lib.rs - Poller::interest (line 262) ... ok
test src/lib.rs - Poller::interest (line 251) ... ok
test src/lib.rs -  (line 16) ... ok
test src/lib.rs - Poller::new (line 168) ... ok
test src/lib.rs - Poller::notify (line 350) ... ok
test src/lib.rs - Poller::remove (line 285) ... ok
test src/lib.rs - Poller::insert (line 189) ... ok
test src/lib.rs - Poller::wait (line 318) ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,10 @@ https://docs.rs/polling)
 Portable interface to epoll, kqueue, and wepoll.
 
 Supported platforms:
-- [epoll](https://en.wikipedia.org/wiki/Epoll): Linux, Android, illumos
+- [epoll](https://en.wikipedia.org/wiki/Epoll): Linux, Android
 - [kqueue](https://en.wikipedia.org/wiki/Kqueue): macOS, iOS, FreeBSD, NetBSD, OpenBSD,
   DragonFly BSD
+- [event ports](https://illumos.org/man/port_create): illumos, Solaris
 - [wepoll](https://github.com/piscisaureus/wepoll): Windows
 
 Polling is done in oneshot mode, which means interest in I/O events needs to be reset after
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,9 +1,10 @@
-//! Portable interface to epoll, kqueue, and wepoll.
+//! Portable interface to epoll, kqueue, event ports, and wepoll.
 //!
 //! Supported platforms:
-//! - [epoll](https://en.wikipedia.org/wiki/Epoll): Linux, Android, illumos
+//! - [epoll](https://en.wikipedia.org/wiki/Epoll): Linux, Android
 //! - [kqueue](https://en.wikipedia.org/wiki/Kqueue): macOS, iOS, FreeBSD, NetBSD, OpenBSD,
 //!   DragonFly BSD
+//! - [event ports](https://illumos.org/man/port_create): illumos, Solaris
 //! - [wepoll](https://github.com/piscisaureus/wepoll): Windows
 //!
 //! Polling is done in oneshot mode, which means interest in I/O events needs to be reset after
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,9 +70,15 @@ macro_rules! syscall {
 }
 
 cfg_if! {
-    if #[cfg(any(target_os = "linux", target_os = "android", target_os = "illumos"))] {
+    if #[cfg(any(target_os = "linux", target_os = "android"))] {
         mod epoll;
         use epoll as sys;
+    } else if #[cfg(any(
+        target_os = "illumos",
+        target_os = "solaris",
+    ))] {
+        mod port;
+        use port as sys;
     } else if #[cfg(any(
         target_os = "macos",
         target_os = "ios",
--- a/src/port.rs
+++ b/src/port.rs
@@ -1,0 +1,214 @@
+//! Bindings to event port (illumos, Solaris).
+
+use std::ptr;
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::io::{self, Read, Write};
+use std::time::Duration;
+use std::os::unix::net::UnixStream;
+use std::usize;
+
+use crate::Event;
+
+/// Interface to epoll.
+#[derive(Debug)]
+pub struct Poller {
+    /// File descriptor for the epoll instance.
+    port_fd: RawFd,
+    /// Read side of a pipe for consuming notifications.
+    read_stream: UnixStream,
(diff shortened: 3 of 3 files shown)


papertigers:
cc @pfmooney @jclulow would appreciate if either of you have time to look this over.

papertigers:
As promised I tested with 

Working proxy with async-std (master branch) + this patched polling crate
```
[root@2c0a4ebb-0d71-4d16-b83b-b849d57238e5 ~/src/proxy-rs]# ./target/debug/proxy-rs -l 10.8.8.109:8080 -r 10.0.1.124:22
2020-08-06 23:13:53 Proxy listening on: 10.8.8.109:8080
2020-08-06 23:14:07 New connection from: 10.0.1.22:59482
2020-08-06 23:14:12 Connection closed: (10.0.1.22:59482 -> 10.0.1.124:22) sent: 3.2 kB received: 4.7 kB (5.1s)
2020-08-06 23:14:13 New connection from: 10.0.1.22:59492
2020-08-06 23:14:14 Connection closed: (10.0.1.22:59492 -> 10.0.1.124:22) sent: 3.0 kB received: 3.7 kB (1.0s)
```

And the passing test suite:
```
[root@2c0a4ebb-0d71-4d16-b83b-b849d57238e5 ~/src/polling]# rustup show
Default host: x86_64-unknown-illumos
rustup home:  /root/.rustup

nightly-x86_64-unknown-illumos (default)
rustc 1.47.0-nightly (22ee68dc5 2020-08-05)
[root@2c0a4ebb-0d71-4d16-b83b-b849d57238e5 ~/src/polling]# cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running target/debug/deps/polling-9a10b7ea2a5dc346

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests polling

running 10 tests
test src/lib.rs - Poller::interest (line 251) ... ok
test src/lib.rs - Poller::interest (line 229) ... ok
test src/lib.rs - Poller::interest (line 262) ... ok
test src/lib.rs - (line 16) ... ok
test src/lib.rs - Poller::interest (line 240) ... ok
test src/lib.rs - Poller::new (line 168) ... ok
test src/lib.rs - Poller::notify (line 350) ... ok
test src/lib.rs - Poller::insert (line 189) ... ok
test src/lib.rs - Poller::remove (line 285) ... ok
test src/lib.rs - Poller::wait (line 318) ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

stjepang:
Nice, thank you so much! :) I wonder, is there a way to test this in CI?

papertigers:
We unfortunately don't have a public CI system out there that supports illumos.  It's definitely something we would like to see happen though. Hopefully we can get something soon because I know various rust crate authors have asked us for the same thing.

stjepang:
I see, that is okay, I think we can get by without CI just fine :)

stjepang:
I'm taking a closer look at this again and I have a few questions. It seems to me that event ports are more similar to kqueue than to epoll. Therefore:

1. Should we ignore ENOENT errors when removing a fd from poller? We don't really know whether the fd is still associated with the port fd (depends on whether an interest is set), so it's best to ignore such errors.

2. When setting an interest in fd, we only set POLLIN and POLLOUT flags. Should we not use the whole set of `read_flags()` and `write_flags()` instead?

3. When setting an interest in fd, we're not removing interests if `readable` or `writable` is `false`. Should we do that like in the kqueue bindings?

papertigers:
I may have misunderstood insert() API.  It looks like the kqueue version just sets the fd to O_NONBLOCK, where as the epoll version sets the fd to O_NONBLOCK and adds the fd to epoll with a events set to 0.

1.  That's a good point.  Event ports always operates in oneshot mode so after something has returned an event it is up to the caller to reregister the fd with port_associate(). I think the man page suggests we will get EBADFD if the fd is not valid when calling port_dissociate().

2. Correct. This was something I asked for clarity on myself. The poll(2) man page says that POLLHUP and POLLERR for example are only returned in the events field and should not be set by the caller when calling port_associate().  The man page refers to these as `events` and `revents`. 
A good example of this is looking at what the HAProxy program does:
https://github.com/haproxy/haproxy/blob/master/src/ev_evports.c#L43-L46
https://github.com/haproxy/haproxy/blob/master/src/ev_evports.c#L241-L247

stjepang:
Regarding 1., reading https://illumos.org/man/3C/port_dissociate makes me believe ENOENT is the error type to look for:

>  The port_dissociate() function will fail if:
>
>       EACCES
>                 The process is not the owner of the association.
>
>
>       ENOENT
>                 The specified object is not associated with the port.

papertigers:
You are totally right! I was looking at the wrong section.  Yea ENOENT is what we want there.
And we can remove the port_associate() call in the intesert function?  Since it doesn't look like we actually need to add it to event ports with no events set.  

Want me to open a PR for this?

stjepang:
Sure! Thanks :)

PR #10, opened 2020-08-27
Check if `epoll_create1` is implemented by its error code

oblique:
If binary is statically linked to libc then `libc::dlsym` will always return `NULL`. This is the case when a target with musl is used.

To fix this we check if error code of `epoll_create1` is `ENOSYS`.

### DIFF
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -16,7 +16,7 @@ pub struct Poller {
     /// File descriptor for the eventfd that produces notifications.
     event_fd: RawFd,
     /// File descriptor for the timerfd that produces timeouts.
-    timer_fd: RawFd,
+    timer_fd: Option<RawFd>,
 }
 
 impl Poller {
@@ -49,17 +49,24 @@ impl Poller {
 
         // Set up eventfd and timerfd.
         let event_fd = syscall!(eventfd(0, libc::EFD_CLOEXEC | libc::EFD_NONBLOCK))?;
-        let timer_fd = syscall!(timerfd_create(
-            libc::CLOCK_MONOTONIC,
-            libc::TFD_CLOEXEC | libc::TFD_NONBLOCK,
-        ))?;
+        let timer_fd = syscall!(syscall(
+            libc::SYS_timerfd_create,
+            libc::CLOCK_MONOTONIC as libc::c_int,
+            (libc::TFD_CLOEXEC | libc::TFD_NONBLOCK) as libc::c_int,
+        ))
+        .map(|fd| fd as libc::c_int)
+        .ok();
+
         let poller = Poller {
             epoll_fd,
             event_fd,
             timer_fd,
         };
+
+        if let Some(timer_fd) = timer_fd {
+            poller.insert(timer_fd)?;
+        }
         poller.insert(event_fd)?;
-        poller.insert(timer_fd)?;
         poller.interest(
             event_fd,
             Event {
@@ -70,7 +77,7 @@ impl Poller {
         )?;
 
         log::trace!(
-            "new: epoll_fd={}, event_fd={}, timer_fd={}",
+            "new: epoll_fd={}, event_fd={}, timer_fd={:?}",
             epoll_fd,
             event_fd,
             timer_fd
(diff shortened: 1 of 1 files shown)


oblique:
I marked it as draft because I think it will produce linkage error with glibc < 2.9. In that case I need to use `libc::syscall`. I will check this and let you know.

oblique:
Ok ready

stjepang:
Nice! So just to confirm - does this work now with glibc < 2.9?

@dignifiedquire This PR looks good to me, but just wanted to check with you because I believe you previously reported errors on old versions of Android that didn't have `epoll_create1`.

oblique:
> Nice! So just to confirm - does this work now with glibc < 2.9?

I didn't find a system to test it but yes. `SYS_epoll_create1` is defined in `libc` crate as `c_long`. And kernel < 2.6.27 will return `ENOSYS` if system call number is invalid. So it does not depend at all in glibc.

stjepang:
Maybe a silly question, but: is `SYS_epoll_create1` present even on unix systems that don't support it? My fear is that this would cause compilation errors due to `libc::SYS_epoll_create1` not being present, but I don't know if that should be a concern.

oblique:
Yes it is present in all Linux systems as a number in libc crate.

stjepang:
Ok great! In that case, I'm ready to merge this PR. :) I'm just gonna wait to see if @dignifiedquire is okay with merging too.

dignifiedquire:
I am quite certain that trying to link to it will result in issues on the android systems I am trying to support. I won’t be close to a computer the next week, so can’t test it myself

oblique:
@dignifiedquire No problem, we can wait. If indeed you have linkage error, paste it here.

src/epoll.rs
```diff
@@ -30,48 +30,47 @@ impl Poller {
...
-                let epoll_create1 = std::mem::transmute::<
-                    *mut libc::c_void,
-                    unsafe extern "C" fn(libc::c_int) -> libc::c_int,
-                >(ptr);
-                match epoll_create1(CLOEXEC) {
-                    -1 => return Err(io::Error::last_os_error()),
-                    fd => fd,
+        //
+        // Use `epoll_create1` with `CLOEXEC`.
+        let epoll_fd = syscall!(syscall(libc::SYS_epoll_create1, CLOEXEC))
```
stjepang, line 35: Is there any point in defining `const CLOEXEC` above if it's only used here? We could just put `libc::EPOLL_CLOEXEC` here, right?

src/epoll.rs
```diff
@@ -151,41 +150,44 @@ impl Poller {
...
-        // Set interest in timerfd.
-        self.interest(
-            self.timer_fd,
-            Event {
-                key: crate::NOTIFY_KEY,
-                readable: true,
-                writable: false,
-            },
-        )?;
+            syscall!(syscall(libc::SYS_timerfd_settime, timer_fd, 0, &new_val, 0))?;
```
stjepang, line 166: Let's be explicit with types to make sure we don't call the syscall with arguments of incorrect types:
```suggestion
            syscall!(syscall(libc::SYS_timerfd_settime, timer_fd, 0 as libc::c_int, &new_val as *const libc::itimerspec, 0 as *mut libc::itimerspec))?;
```

src/epoll.rs
oblique: correct

src/epoll.rs
```diff
@@ -16,62 +16,57 @@ pub struct Poller {
...
-                // Use `epoll_create1` with `CLOEXEC`.
-                let epoll_create1 = std::mem::transmute::<
-                    *mut libc::c_void,
-                    unsafe extern "C" fn(libc::c_int) -> libc::c_int,
-                >(ptr);
-                match epoll_create1(CLOEXEC) {
-                    -1 => return Err(io::Error::last_os_error()),
-                    fd => fd,
+        //
+        // Use `epoll_create1` with `CLOEXEC`.
```
stjepang, line 27:
```suggestion
        // Use `epoll_create1` with `EPOLLCLOEXEC`.
```

src/epoll.rs
```diff
@@ -16,62 +16,57 @@ pub struct Poller {
...
+        let epoll_fd = syscall!(syscall(
+            libc::SYS_epoll_create1,
+            libc::EPOLL_CLOEXEC as libc::c_int
+        ))
+        .map(|fd| fd as libc::c_int)
+        .or_else(|e| {
+            match e.raw_os_error() {
+                Some(libc::ENOSYS) => {
+                    // If `epoll_create1` is not implemented, use `epoll_create`
+                    // and manually set `CLOEXEC`.
```
stjepang, line 37:
```suggestion
                    // and manually set `FD_CLOEXEC`.
```

stjepang (approved):
We got a green light from @dignifiedquire so this is ready to merge. Thanks everyone!

PR #14, opened 2020-09-06
Fix compilation on x32

sunfishcode:
Fix compilation on x86_64-unknown-linux-gnux32, where
libc's timespec's `tv_nsec` is [defined as `i64`]. See also
[this bug] bug for further discussion.

[defined as `i64`]: https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs#L61
[this bug]: https://sourceware.org/bugzilla/show_bug.cgi?id=16437


stjepang (approved):
Thank you!

stjepang:
Published v0.1.9

PR #15, opened 2020-09-16
remove `libc` dependency on windows

JayceFayne:
Like in [async-io](https://github.com/stjepang/async-io/blob/master/Cargo.toml#L26) we only need `libc` on unix :smile:


stjepang (approved):
Thank you, really appreciate it! :)

PR #17, opened 2020-09-30
Refactor Poller.add and Poller.interest

YorickPeterse:
This MR addresses the discussion in https://github.com/stjepang/polling/issues/16:

1. A `Poller` no longer makes file descriptors non-blocking, and instead leaves this up to the user
1. `Poller.insert()` and `Poller.interest()` have been replaced with `Poller.add()`, `Poller.update()`, and `Poller.delete()`

### DIFF
--- a/examples/two-listeners.rs
+++ b/examples/two-listeners.rs
@@ -11,11 +11,9 @@ fn main() -> io::Result<()> {
     l2.set_nonblocking(true)?;
 
     let poller = Poller::new()?;
-    poller.insert(&l1)?;
-    poller.insert(&l2)?;
 
-    poller.interest(&l1, Event::readable(1))?;
-    poller.interest(&l2, Event::readable(2))?;
+    poller.add(&l1, Event::readable(1))?;
+    poller.add(&l2, Event::readable(2))?;
 
     let mut events = Vec::new();
     loop {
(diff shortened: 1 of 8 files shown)


YorickPeterse:
This is still a draft, so item number 2 still needs to be addressed. I'll do so tomorrow.

examples/two-listeners.rs
```diff
@@ -7,6 +7,12 @@ fn main() -> io::Result<()> {
     let l1 = TcpListener::bind("127.0.0.1:8001")?;
     let l2 = TcpListener::bind("127.0.0.1:8002")?;
 
+    l1.set_nonblocking(true)
+        .expect("Failed to make the socket non-blocking");
```
stjepang, line 11: Let's just put a `?` at the end :)

stjepang:
I'm thinking we should pick one of the following two naming schemes:

1. `insert()`/`update()`/`remove()` - this follows the naming conventions in `std::collections`.
2. `add()`/`modify()`/`delete()` - this follows the naming conventions of epoll and kqueue (API and wording in manual pages).

What do you think? Any preferences?

YorickPeterse:
@stjepang I don't have any strong opinions, but I think sticking with the epoll/kqueue naming scheme is best.

stjepang:
Sounds good! Let's go with that then.

YorickPeterse:
@stjepang There appear to be build failures because async-io is not compatible with these changes, which is to be expected. Is there anything I need to do about that in this pull request, or do we need to tackle that separately?

stjepang:
You can comment that test out for the time being.

src/lib.rs
```diff
@@ -100,6 +101,17 @@ cfg_if! {
...
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "the key is not allowed to be `usize::MAX`",
+            ));
+        }
+    }};
+}
+
 /// Key associated with notifications.
 const NOTIFY_KEY: usize = std::usize::MAX;
```
stjepang, line 116: I think this needs a `#[cfg(...)]` to get around the warnings.

YorickPeterse: @stjepang Which warnings are you talking about? I can't seem to find any in the (horrible) GitHub checks UI.

stjepang (approved):

PR #18, opened 2020-10-02
Try fixing sigsegv

stjepang:


stjepang:
@YorickPeterse There's a problem. If I use this, everything works fine:

```toml
wepoll-sys = { git = "https://github.com/stjepang/wepoll-sys-stjepang.git", package = "wepoll-sys-stjepang" }
```

But when I switch to this, I get SIGSEGV in tests:

```toml
wepoll-sys = "2.0.0"
```

The CI failure is here: https://github.com/stjepang/polling/runs/1199877088#step:10:27

I noticed that `wepoll.c` is the same on the master branches of `wepoll-sys` and `wepoll-sys-stjepang`. However, the file is different in published `wepoll-sys` v2.0.0. Perhaps something went wrong while publishing the crate?

YorickPeterse:
@stjepang You need to use version 3.0.0 of wepoll-sys, version 2.0.0 is quite old :smile:

stjepang:
Ah, my bad! Thanks, this should work now :)

PR #26, opened 2020-12-17
Support VxWorks, Fuchsia and other Unix systems by using poll

### LINKED ISSUE #20: Add POSIX poll-based implementation
This crate does not compile on Haiku. libuv uses `poll` on Haiku, would be nice to also add `poll`-based implementation to this crate as a baseline for operating systems that do not support faster interfaces.

Koxiaet:
Closes #20.

Notes:
* I had to add the `extra_traits` feature to `libc` to `#[derive(Debug)]` on the polling implementation. This can be done manually to avoid activating the feature, if that's something you care about.
* I had to `#[derive(Clone, Copy)]` on `Event`. It seems to be fine as `Event`'s interface is already fully exposed.
* I have tested this on Linux by disabling `epoll` and I have checked it compiles on Fuchsia (VxWorks and Haiku aren't supported by Rustup).

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -3,8 +3,9 @@
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::io;
-use std::sync::Mutex;
-use std::time::Duration;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::{Condvar, Mutex};
+use std::time::{Duration, Instant};
 
 // std::os::unix doesn't exist on Fuchsia
 use libc::c_int as RawFd;
@@ -19,11 +20,27 @@ const REMOVE_FD: RawFd = -2;
 pub struct Poller {
     /// File descriptors to poll.
     fds: Mutex<Fds>,
+
     /// The file descriptor of the read half of the notify pipe. This is also stored as the first
     /// file descriptor in `fds.poll_fds`.
     notify_read: RawFd,
     /// The file descriptor of the write half of the notify pipe.
+    ///
+    /// Data is written to this to wake up the current instance of `wait`, which can occur when the
+    /// user notifies it (in which case `notified` would have been set) or when an operation needs
+    /// to occur (in which case `waiting_operations` would have been incremented).
     notify_write: RawFd,
+
+    /// The number of operations (`add`, `modify` or `delete`) that are currently waiting on the
+    /// mutex to become free. When this is nonzero, `wait` must be suspended until it reaches zero
+    /// again.
+    waiting_operations: AtomicUsize,
+    /// The condition variable that gets notified when `waiting_operations` reaches zero. This is
+    /// used with the `fds` mutex.
+    operations_complete: Condvar,
+
+    /// Whether `wait` has been notified by the user.
+    notified: AtomicBool,
 }
 
 /// The file descriptors to poll in a `Poller`.
@@ -76,6 +93,9 @@ impl Poller {
             }),
             notify_read: notify_pipe[0],
             notify_write: notify_pipe[1],
+            waiting_operations: AtomicUsize::new(0),
+            operations_complete: Condvar::new(),
+            notified: AtomicBool::new(false),
         })
     }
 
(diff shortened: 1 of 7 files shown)


stjepang:
Some quick questions before I do a deeper dive into the code...

Suppose a thread is blocked on `wait()`. Another thread then calls `add()` and the added FD is immediately ready.

1. Is it ensured that `wait()` doesn't hold a mutex lock that blocks `add()` forever?
2. Is it ensured that `add()` wakes up the caller of `wait()` even though the added FD is not in its poll set?

Koxiaet:
Thanks for pointing that out, I didn't consider those cases - I'll convert this to a draft.

Koxiaet:
I've changed the implementation to allow `add`, `modify` and `delete` to

Koxiaet:
Ok, I think I've handled those cases now. Any call to `add`, `modify` or `delete` will interrupt the thread that's calling `poll` and send it into a condvar loop which it will exit if the user calls `.notify()` or all the operations complete.

src/poll.rs
```diff
@@ -0,0 +1,357 @@
...
+    loop {
+        // Convert the timeout to milliseconds.
+        let timeout_ms = deadline
+            .map(|deadline| {
+                let timeout = deadline.saturating_duration_since(Instant::now());
+
+                // Round up to a whole millisecond.
+                let mut ms = timeout.as_millis().try_into().unwrap_or(std::u64::MAX);
+                if Duration::from_millis(ms) < timeout {
+                    ms += 1;
```
zseri, line 344: it is not clear that this will never overflow.

src/poll.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// The list of `pollfds` taken by poll.
+    ///
+    /// The first file descriptor is always present and is used to notify the poller. It is also
+    /// stored in `notify_read`.
+    ///
+    /// If the fd stored in here is `REMOVE_FD`, it should be removed.
+    poll_fds: Vec<libc::pollfd>,
+    /// The map of each file descriptor to data associated with it. This does not include the file
+    /// descriptors `notify_read` or `notify_write`.
+    fd_data: HashMap<RawFd, FdData>,
```
zseri, line 59: This seems to just implement an separately-ordered, hashmap. Why don't we pull in something like [`indexmap`](https://github.com/bluss/indexmap) for this, or do we require some feature it doesn't have? (if true, which?)

Kestrer: I'm averse to adding a dependency for this implementation, mostly because it would require a really long `cfg` in the Cargo.toml :smile:. However, this comment did make me realize that I can make the implementation more efficient and simpler by using `swap_remove` instead of marking fds as removed and later removing them.

zseri:
We use
```toml
[target."cfg(unix)".dependencies]
libc = { version = "0.2.77", features = ["extra_traits"] }
```
but
```rust
    } else if #[cfg(any(
        target_os = "vxworks",
        target_os = "fuchsia",
        unix,
    ))] {
        mod poll;
        use poll as sys;
```
and
```rust
// std::os::unix doesn't exist on Fuchsia
use libc::c_int as RawFd;
```

This is a mismatch between cfg-targets and should be fixed.
btw. if fixed, this would mean the justification in "I'm averse to adding a dependency for this implementation, mostly because it would require a really long cfg in the Cargo.toml" would no longer apply....

Kestrer:
Ah right, the reason Fuchsia was compiling before is that it hasn't actually been changed to a `target_family` of `None` yet (it's still Unix).

> if fixed, this would mean the justification in "I'm averse to adding a dependency for this implementation, mostly because it would require a really long cfg in the Cargo.toml" would no longer apply....

It still applies, because the actual `cfg` required there would be `cfg(any(target_os = "fuchsia", target_os = "vxworks", all(unix, not(any(target_os = "linux", target_os = "android", target_os = "illumos", target_os = "solaris", target_os = "macos", target_os = "ios", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd", target_os = "dragonfly")))))`.

link2xt:
I have tested in on Haiku, works for me. Rustup indeed does not support Haiku, but rust can be installed with `pkgman install rust_bin`.

taiki-e:
I haven't reviewed this PR yet, but in mio, there was a discussion on `poll`'s problems in the past: https://github.com/tokio-rs/mio/issues/1472#issuecomment-796816730 
How does this PR deal with such problems?

link2xt:
> I haven't reviewed this PR yet, but in mio, (IIUC) there was a discussion on `poll`'s problems in the past: [tokio-rs/mio#1472 (comment)](https://github.com/tokio-rs/mio/issues/1472#issuecomment-796816730)
> How does this PR deal with such problems?

By storing all file descriptors inside a `Poller`, it's the first field of type `Mutex<Fds>`. Looks like `polling` does not have problems supporting `poll`.

taiki-e (changes requested):
This looks good to me. Thanks for working on this!
(I'm not familiar with the platforms mentioned here, but I have tweaked this patch a bit and tested polling, async-io, and async-net on macOS.)

- Could you update "supported platforms" docs on readme?
- Could you add Fuchsia to [cross workflow on CI](https://github.com/smol-rs/polling/blob/952fccb2f56fbc6eaf7d6d80de5e6704a76f4c21/.github/workflows/cross.yaml)? Probably something like:

    ```yaml
    - name: Fuchsia
      if: startsWith(matrix.os, 'ubuntu')
      run: cross build --target x86_64-fuchsia
    ```

taiki-e:
Also, would it be possible to add tests for the cases mentioned in https://github.com/smol-rs/polling/pull/26#issuecomment-747656221 by Stjepan? I don't think tests for such a case currently exist.

> Suppose a thread is blocked on `wait()`. Another thread then calls `add()` and the added FD is immediately ready.
> 
> 1. Is it ensured that `wait()` doesn't hold a mutex lock that blocks `add()` forever?
> 2. Is it ensured that `add()` wakes up the caller of `wait()` even though the added FD is not in its poll set?

src/poll.rs
```diff
@@ -0,0 +1,377 @@
...
+                // Round up to a whole millisecond.
+                let mut ms = timeout.as_millis().try_into().unwrap_or(std::u64::MAX);
+                if Duration::from_millis(ms) < timeout {
+                    ms = ms.saturating_add(1);
+                }
+                ms.try_into().unwrap_or(std::i32::MAX)
+            })
+            .unwrap_or(-1);
+
+        match syscall!(poll(fds.as_mut_ptr(), fds.len() as u64, timeout_ms,)) {
```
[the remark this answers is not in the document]
taiki-e:
```suggestion
        match syscall!(poll(fds.as_mut_ptr() as *mut libc::pollfd, fds.len() as libc::nfds_t, timeout_ms,)) {
```

Kestrer:
OK, I have done everything suggested 😊

Hopefully CI will pass...

taiki-e:
Published in v2.2.0.

Issue #35, opened 2021-05-25
Reducing licenses used in dependency tree


aclysma:
Hello, a few of us are working on an OSS rust crate called `distill`. (https://github.com/amethyst/distill) We were looking to transition from tokio to some of the other smaller async crates like async-io and async-net.

`polling` is one of the dependencies, and it depends on `wepoll-sys`. However, the license (MPL) is technically a copyleft license and we want to avoid copyleft licenses in our dependency tree.

I know licenses can be a controversial topic, and I’m not looking to debate the merits of any particular license. However, it appears that other than this one dependency, the entire dependency tree of async-std is free from any form of copyleft licenses. 

I’ve contacted the author of the crate, but I think they prefer to keep the license the same, and I fully respect their wishes!

Thankfully, it’s not much work to create new bindings for wepoll-sys. I’ve already done this and plan to release it under Apache-2.0 OR MIT OR BSD-2-Clause (wepoll itself is BSD-2-Clause). Functionally, the only thing missing was a patch from @stjepang that was discussed here: https://github.com/piscisaureus/wepoll/pull/20. I’m not sure what it is for or if it’s still necessary. I’m also not sure if @stjepang is willing to have it included, or if I would have to reimplement it if it is still needed.

(I'm equally happy for these new bindings to be owned by someone in the smol-rs org, if someone would like to take ownership of them)
 
I wanted to ask what other people think before going further. Again, I know there is potential for controversy with this topic, so I’m very sorry if it leads to any sort of awkward discussion. My angle here is just trying to reduce the number of licenses a user has to accept, or a company’s legal team has to clear, in order to use this crate, either directly, or indirectly via other ecosystem crates like async-std.

zeenix:
> I’m not sure what it is for or if it’s still necessary. I’m also not sure if @stjepang is willing to have it included, or if I would have to reimplement it if it is still needed.

@stjepang is no longer working on smol-rs or anything related so they're very unlikely to be interested. Given that `polling` isn't (yet) dependent on the changes in that open PR, I think it's safe to assume that that is not a blocker (for now at least).

> (I'm equally happy for these new bindings to be owned by someone in the smol-rs org, if someone would like to take ownership of them)

Given all the smol-rs crates have a very liberal license (including `polling` itself), at least I would be more than willing to accept this change but you'll need to commit to maintaining your wepoll replacement crate I'm afraid and provide the PR for it.

@smol-rs/admins WDYT?

aclysma:
[...]

PR #36, opened 2021-05-26
Switch wepoll-sys to wepoll-ffi to reduce licenses used in dependency tree (discussed in #35)

aclysma:
This PR is based on discussion in #35 and removes the only MPL-licensed dependency from the async-std dependency tree. All remaining licenses in it are very permissive (i.e. Apache-2.0/MIT or similar)

The new FFI bindings crate has been published to crates.io:
https://crates.io/crates/wepoll-ffi
https://github.com/aclysma/wepoll-ffi 

The main differences AFAIK from wepoll-sys:
 - bindgen by default named the "u64" fields in C as "u64_". I personally prefer sticking to bindgen's defaults unless there is reason not to (i.e. unusually awkward ergonomics).
 - This would effectively remove the patch discussed in https://github.com/piscisaureus/wepoll/pull/20 (meaning we would be using vanilla wepoll 1.5.8)

As discussed in #35, I would lean towards not including the patch as it has not been upstreamed to wepoll itself. But I'm happy to add it if we know it's needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ log = "0.4.11"
 libc = "0.2.77"
 
 [target.'cfg(windows)'.dependencies]
-wepoll-ffi = "0.1.0"
+wepoll-ffi = { version = "0.1.1", features = ["null-overlapped-wakeups-patch"] }
 winapi = { version = "0.3.9", features = ["ioapiset", "winsock2"] }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ log = "0.4.11"
 libc = "0.2.77"
 
 [target.'cfg(windows)'.dependencies]
-wepoll-ffi = { version = "0.1.1", features = ["null-overlapped-wakeups-patch"] }
+wepoll-ffi = { version = "0.1.2", features = ["null-overlapped-wakeups-patch"] }
 winapi = { version = "0.3.9", features = ["ioapiset", "winsock2"] }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ log = "0.4.11"
 libc = "0.2.77"
 
 [target.'cfg(windows)'.dependencies]
-wepoll-sys = "3.0.0"
+wepoll-ffi = "0.1.0"
 winapi = { version = "0.3.9", features = ["ioapiset", "winsock2"] }
 
 [dev-dependencies]
--- a/src/wepoll.rs
+++ b/src/wepoll.rs
@@ -7,7 +7,7 @@ use std::ptr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::{Duration, Instant};
 
-use wepoll_sys as we;
+use wepoll_ffi as we;
 use winapi::ctypes;
 
 use crate::Event;
@@ -149,7 +149,7 @@ impl Poller {
             }
             we::epoll_event {
                 events: flags as u32,
-                data: we::epoll_data { u64: ev.key as u64 },
+                data: we::epoll_data { u64_: ev.key as u64 },
             }
         });
         wepoll!(epoll_ctl(
@@ -192,7 +192,7 @@ impl Events {
     pub fn new() -> Events {
         let ev = we::epoll_event {
             events: 0,
-            data: we::epoll_data { u64: 0 },
+            data: we::epoll_data { u64_: 0 },
         };
         Events {
             list: vec![ev; 1000].into_boxed_slice(),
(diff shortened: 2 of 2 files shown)


aclysma:
I now understand why the patch is necessary. Polling calls `PostQueuedCompletionStatus` passing a null lpOverlapped param to implement notify(). wepoll assumes lpOverlapped will never be null. I can add a patch that handles this that is optionally enabled with a #define, and expose a feature on the crate to enable it. This way polling can have the patch, but other users can get vanilla wepoll behavior by default. Does this sound like a good way to solve it?

aclysma:
Diff of changes to wepoll here: https://github.com/aclysma/wepoll/commit/3cf5cf446b9b30bd98803cbdc7c7ab8a791789c9

It is enabled by using the null-overlapped-wakeups-patch crate feature on wepoll-ffi, which defines NULL_OVERLAPPED_WAKEUPS_PATCH.

Worth noting that the tests do find the crash without this feature/patch, and once added, all tests pass.

zeenix:
I feel like I'm very unqualified for reviewing this PR as I've not ideas about windows and its systems. But given that the patches seem very simple and straight-forward enough, all the tests pass and @aclysma has done a great job with communicating their goals, with which I agree, I'll merge this tomorrow unless someone objects. (@smol-rs/admins ?).

aclysma:
I made one more small modification to that patch: https://github.com/aclysma/wepoll/commit/48520c99d93e8c2b5b0c067e747c0eef3d316522 I actually meant to make this constant be something *other* than -1, and for whatever reason I didn't connect that 0xFFFFFFFF is just another way to write -1.

I also want to fully explain why this patch is necessary and how it works to save trouble for anyone looking at this in the future :)

 * Say some thread calls epoll_wait. We will enter a loop in port_wait (https://github.com/aclysma/wepoll/blob/48520c99d93e8c2b5b0c067e747c0eef3d316522/wepoll.c#L1311-L1315). This will repeatedly call port__poll, which call the windows function GetQueuedCompletionStatusEx until some event is received.
 * Now say we want to force the thread to wake up.. we do this with this line of code in the `polling` crate:
https://github.com/smol-rs/polling/blob/1019f90247a08b5e2a30b11d54792e5354c3f6a1/src/wepoll.rs#L127-L134
 * The last parameter is lpOverlapped, the windows documentation for this function is here (https://docs.microsoft.com/en-us/windows/win32/fileio/postqueuedcompletionstatus) - the docs describe **lpOverlapped** as "The value to be returned through the lpOverlapped parameter of the GetQueuedCompletionStatus function."
     * **We are passing null for lpOverlapped**!
 * As previously mentioned, the waiting thread called GetQueuedCompletionStatusEx here (https://github.com/aclysma/wepoll/blob/48520c99d93e8c2b5b0c067e747c0eef3d316522/wepoll.c#L1254-L1259)
     * (docs: https://docs.microsoft.com/en-us/windows/win32/fileio/getqueuedcompletionstatusex-func).
 * The second parameter is an out parameter, a list of OVERLAPPED_ENTRY structs (https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped_entry)
 * Each struct has a field **lpOverlapped**, the docs describe it as "Receives the address of the OVERLAPPED structure that was specified when the completed I/O operation was started."
     * **If the event we receive is from our PostQueuedCompletionStatus call, lpOverlapped will be null** because we passed null when we called PostQueuedCompletionStatus
 * We call port__feed_events, passing this list of structs that has a null lpOverlapped. Original code for that function is here https://github.com/piscisaureus/wepoll/blob/0598a791bf9cbbf480793d778930fc635b044980/wepoll.c#L1191
     * It passes the null lpOverlapped into sock_feed_event(), which will dereference the null pointer here: https://github.com/piscisaureus/wepoll/blob/0598a791bf9cbbf480793d778930fc635b044980/wepoll.c#L1849
     * **So if notify() is used with unpatched wepoll, we will trigger undefined behavior and very likely crash.**
 * The patched code checks for null lpOverlapped:
     * If we receive a null lpOverlapped event **AND** other events, we just drop the event with null lpOverlapped. The other events are already going to wake the thread, so we don't need to do anything special.
     * If we receive a null lpOverlapped event, but there are no other events, we need to do something to wake the thread. (This case: https://github.com/aclysma/wepoll/blob/48520c99d93e8c2b5b0c067e747c0eef3d316522/wepoll.c#L1231)
         * We return a new error code to indicate to the waiting thread that it should break out of the wait
         * This error code is caught in the loop here. We change the result to be 0 (non-error, no events received) and break out of the loop, which wakes the thread: https://github.com/aclysma/wepoll/blob/48520c99d93e8c2b5b0c067e747c0eef3d316522/wepoll.c#L1319

I changed the error code to be something other than -1 because port__poll could already return -1 here. This ought to be passed through as an error back to the original epoll_wait call. If we used -1, then that error case would be silenced: https://github.com/aclysma/wepoll/blob/48520c99d93e8c2b5b0c067e747c0eef3d316522/wepoll.c#L1248

PR #39, opened 2022-08-17
Expose raw handles for the `Poller`

notgull:
Most of the current `Poller` instances are wrappers around either a `RawFd` or a `RawHandle`. For these platforms, an `AsRawFd` or `AsRawHandle` implementation is implemented on `Poller`. Then, because I have absolutely no self control, I also went ahead and conditionally implemented `AsFd` and `AsHandle` as well on conditionally proper Rust versions through `autocfg`. See sunfishcode/io-lifetimes#38 for more information.

This is part of my attempt at getting smol-rs/async-io#39 to work

### DIFF
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -2,9 +2,12 @@
 
 use std::convert::TryInto;
 use std::io;
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsRawFd, RawFd};
 use std::ptr;
 use std::time::Duration;
+
+#[cfg(not(polling_no_io_safety))]
+use std::os::unix::io::{AsFd, BorrowedFd};
 
 use crate::Event;
 
@@ -222,6 +225,20 @@ impl Poller {
                 .unwrap_or(ptr::null_mut()),
         ))?;
         Ok(())
+    }
+}
+
+impl AsRawFd for Poller {
+    fn as_raw_fd(&self) -> RawFd {
+        self.epoll_fd
+    }
+}
+
+#[cfg(not(polling_no_io_safety))]
+impl AsFd for Poller {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        // SAFETY: lifetime is bound by "self"
+        unsafe { BorrowedFd::borrow_raw(self.as_raw_fd()) }
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,9 @@ std = []
 cfg-if = "1"
 log = "0.4.11"
 
+[build-dependencies]
+autocfg = "1"
+
 [target.'cfg(any(unix, target_os = "fuchsia", target_os = "vxworks"))'.dependencies]
 libc = "0.2.77"
 
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,16 @@
+fn main() {
+    let cfg = match autocfg::AutoCfg::new() {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            println!(
+                "cargo:warning=polling: failed to detect compiler features: {}",
+                e
+            );
+            return;
+        }
+    };
+
+    if !cfg.probe_rustc_version(1, 63) {
+        autocfg::emit("polling_no_io_safety");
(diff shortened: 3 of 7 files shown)


src/epoll.rs
```diff
@@ -225,6 +228,20 @@ impl Poller {
     }
 }
 
+impl AsRawFd for Poller {
+    fn as_raw_fd(&self) -> RawFd {
+        self.epoll_fd
+    }
+}
+
+#[cfg(not(polling_no_io_safety))]
```
zeenix, line 237: nit: double-negative makes it harder to read (I first misread it). I'd suggest:

```rust
#[cfg(polling_io_safety)]
```

(I also wouldn't use `polling_` prefix but no strong feelings there so just a suggestion :smile:)

zeenix (changes requested):
LGTM otherwise.

src/epoll.rs
taiki-e: This is intentional, see https://github.com/smol-rs/async-io/pull/84#discussion_r948085917

taiki-e: Also, [this comment](https://github.com/tokio-rs/tokio/blob/4b1c4801b1383800932141d0f6508d5b3003323e/tokio/build.rs#L74-L78) describes another case where `no_*` cfg is reasonable.

zeenix: > This is intentional, see https://github.com/smol-rs/async-io/pull/84#discussion_r948085917 for the context.

Thanks but why do we want to assume latest rust if build script is not run? Also, why is build script not run?

I'd be a least good to have [this comment](https://github.com/tokio-rs/slab/blob/817a2ec227e819f308e8e4d0487084940b4ef831/build.rs#L15-L20) here as well so it's clear to the reader, not just me as a reviewer (after asking for it :)).

notgull: I added the comment. There are non-`cargo` build tools that don't run `build.rs` for various reasons. For instance, I can't remember which project it is, but they build some Rust code in a C/C++ codebase by running `rustc` directly through a Makefile. 

In this case, since we have no versioning information available to us, we make the (reasonable) assumption that we're just running the latest stable version. Since `build.rs` is never run, `polling_no_io_safety` is never set.

zeenix: > I added the comment. There are non-`cargo` build tools that don't run `build.rs` for various reasons. For instance, I can't remember which project it is, but they build some Rust code in a C/C++ codebase by running `rustc` directly through a Makefile.

Thanks for explaining! This a Cargo-based project and I really don't see why we should sacrifice the readability of this project for possible users that use cargo-based projects w/o ensuring their build system is actually doing everything cargo-based project expects. However, now that you've added the comment, it's nothing I'd block your good work for. :)

zeenix (approved):

taiki-e:
> This a Cargo-based project and I really don't see why we should sacrifice the readability of this project for possible users that use cargo-based projects w/o ensuring their build system is actually doing everything cargo-based project expects.

AFAIK, non-cargo build systems such as [Bazel](https://github.com/bazelbuild/bazel/), [Buck](https://github.com/facebook/buck), and [GN](https://gn.googlesource.com/gn/) have helpers (e.g., [cargo-raze](https://github.com/google/cargo-raze), [cargo-gnaw](https://gn.googlesource.com/cargo-gnaw)) to create manifests for that build system based on Cargo.toml and can support cargo-based projects. However, running arbitrary other rust code (autocfg) at build time may or may not be simple. Therefore, it is helpful to have reasonable defaults in case the build script is not run. (AFAIK, this was first proposed in https://github.com/serde-rs/serde/issues/1593.)

I maintain several projects that sometimes receive feedback from fuchsia team (using GN, IIRC), so I usually prefer to use negative cfgs so that it works without friction in their code base. (Also, it seems [some smol-rs projects are being used in fuchsia](https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/third_party/rust_crates/vendor/).)

zeenix:
@taiki-e Thanks so much for indulging and educating me. Much appreciated! :+1:

PR #40, opened 2022-08-21
Bump MSRV to 1.47, reduce size of sys::Events

### LINKED ISSUE #19: Smaller Events representation
With the array support that was recently added, smol could safely use a `Box<[platform::event; 1024]>` for the `Events` implementation. It'd bring the MSRV up to 1.47 though, which I'm guessing isn't ideal. Does the crate have an MSRV policy?

taiki-e:
`libc` plans to raise their MSRV to 1.47 (https://github.com/rust-lang/libc/pull/2845), and some crates of smol-rs also require 1.47 since async-task 4.3.0 (https://github.com/smol-rs/async-task/pull/30). It seems to be a reasonable time to do #19.

This also reduces the size of `sys::Events` (closes #19).


src/kqueue.rs
```diff
@@ -221,7 +221,7 @@ impl Events {
             data: 0,
             udata: 0 as _,
         };
-        let list = vec![ev; 1000].into_boxed_slice();
```
taiki-e, line 224: I'm not sure why the old code used 1000 instead of 1024.

src/epoll.rs
```diff
@@ -289,7 +289,7 @@ impl Events {
     /// Creates an empty list.
     pub fn new() -> Events {
         let ev = libc::epoll_event { events: 0, u64: 0 };
-        let list = vec![ev; 1000].into_boxed_slice();
+        let list = Box::new([ev; 1024]);
```
taiki-e, line 292: I didn't do it on this patch, but calling alloc_zeroed via unsafe code is probably best from a point of view of performance.
https://godbolt.org/z/T1e8ETv61

zeenix (approved):

src/kqueue.rs
notgull: It would be a mis-optimization, if I had to guess.

notgull (approved):
Looks good to me!

PR #43, opened 2022-10-23
Fix build error on solarish

### LINKED ISSUE #42: Use of non-existent BorrowFd::new() introduced with 2.3.0
All attempts to build version 2.3.0 on illumos fails with the following error:

```
error[E0599]: no function or associated item named `new` found for struct `BorrowedFd` in the current scope
   --> src/port.rs:159:30
    |
159 |         unsafe { BorrowedFd::new(self.port_fd) }
    |                              ^^^ function or associated item not found in `BorrowedFd<'_>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `polling` due to previous error
```

It appears `src/port.rs` is only included for target_os = "illumos" || "solaris", but [BorrowedFd](https://doc.rust-lang.org/stable/std/os/unix/io/struct.BorrowedFd.html) does not have a `new()` method on any platform. I guess calling `borrowed_raw()` was what was is intended, as it is the only applicable method of the struct? I haven't dug into the issue and this failure is several dependencies deep from the crate I actually use, so I might be very wrong in that assumption.

taiki-e:
This fixes the build error on solarish (illumos, Solaris)

This also re-enables the build for illumos on CI.

Fixes #42.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,15 +69,16 @@ jobs:
     - name: Linux x32
       if: startsWith(matrix.os, 'ubuntu')
       run: cross check --target x86_64-unknown-linux-gnux32
-    - name: Add fuchsia target
-      if: startsWith(matrix.os, 'ubuntu')
-      run: rustup target add x86_64-fuchsia
     - name: Fuchsia
       if: startsWith(matrix.os, 'ubuntu')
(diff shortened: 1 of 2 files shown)


notgull (approved):
Looks like this was my mistake :-/ Anyways, looks good to me!

PR #45, opened 2022-11-20
Relax bound on blanket impl for Source

ybh1998:
The current definition in this library is not friendly for dynamic use like `dyn polling::Source`, because the blanket implementation of `Source` for `&T` (instead of `T`). Consider the following code:
https://github.com/smol-rs/polling/blob/ca52490c9d3638eceaf2f4b96ba56eb23953a7ea/src/lib.rs#L539-L543

The current implementation is effectively accepting a double reference because there are references in both the receiver type and `&T`, and it does not align with other traits that we might also be interested to have (e.g., `Read`, `Write`). Consider the following attempt to have a dynamic `Source` which is also `Read`able and `Write`able:

```rust
pub trait ReadWriteSource: Read + Write + polling::Source {}
impl<T: Read + Write + polling::Source> ReadWriteSource {}
```

Unfortunately, (for example) `TcpStream` does not implement `ReadWriteSource` because it does not implement `Source`.

With the proposed changes, types `TcpStream` will properly implement `ReadWriteSource`. A platform-agnostic return type `RawSource` for `Source::raw` is also added to help if the users want to implement `polling::Source` by themselves.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -240,7 +240,7 @@ impl Poller {
     /// poller.add(&source, Event::all(key))?;
     /// # std::io::Result::Ok(())
     /// ```
-    pub fn add(&self, source: impl Source, interest: Event) -> io::Result<()> {
+    pub fn add<S: Source + ?Sized>(&self, source: &S, interest: Event) -> io::Result<()> {
         if interest.key == NOTIFY_KEY {
             return Err(io::Error::new(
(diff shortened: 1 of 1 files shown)


notgull:
This PR has been displaced by the merge of #123 changing the concept of the `Source` trait. Feel free to re-open if this is updated for the new codebase.

Thanks anyways!

Issue #52, opened 2022-11-30
Poller fails to initialize on old Wine


notgull:
When running this code on Wine (`wine-6.0.3 (Ubuntu 6.0.3~repack-1)`):

```rust
use polling::Poller;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // Create a poller.
    let poller = Poller::new()?;
    
    Ok(())
}
```

It yields this error:

```
Error: Os { code: 10045, kind: Uncategorized, message: "OS Error 10045 (FormatMessageW() returned error 317)" }
```

Error 10045 is [`WSAEOPNOTSUPP`](https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2?redirectedfrom=MSDN#WSAEOPNOTSUPP). It looks like this crate suffers from tokio-rs/mio#1444 as well.

`mio` dealt with this issue by not supporting Wine at all. However, `libuv` deals with this using an implementation that spawns a thread pool that runs `select` for every registered listener. This strategy is significantly slower than `wepoll` but is better than crashing. I feel that we can go in either direction here.

Potentially related: async-rs/async-std#943

notgull:
With #54 being merged, should we close this issue? If desired, I could probably implement a thread-pool based solution on top of `blocking` and `select` to support older versions of Wine. However, as Debian and Ubuntu eventually migrate to newer versions of Wine that support Wepoll, that backend would eventually become obsolete. It may just be in our best interest to explicitly not support older Wines.

taiki-e:
Yeah, given that it works with the latest wine and that it is relatively easy to install the latest wine from winehq, I think it is not very worth it to implement a new backend for old wine support. (The situation is different from libuv, which added fallback before wine supported them.)

PR #53, opened 2022-11-30
Test x86_64-pc-windows-gnu with Wine on CI

taiki-e:
This adds test for x86_64-pc-windows-gnu with Wine 7.13 on CI.

This also documents that Wine 7.13+ is supported.

FYI, in 6.11-7.12 the following error occurs:

```
Error: Os { code: 66, kind: Uncategorized, message: "Bad device type." }
```

And, in pre-6.11 the following error occurs:

```
Error: Os { code: 10045, kind: Uncategorized, message: "OS Error 10045 (FormatMessageW() returned error 317)" }
```

cc #52
cc @notgull

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,18 @@ jobs:
         rustup target add x86_64-unknown-illumos
         cargo build --target x86_64-unknown-illumos
 
+  wine:
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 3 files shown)


notgull (approved):
Looks good to me!

I also think that we should emit some kind of log warning if we detect an old version of Wine running. I can file a PR for that later today.

PR #54, opened 2022-11-30
Add error message for Wepoll

notgull:
This PR adds an error message for initialization of Wepoll that references Wine or an old Windows version. This should help users trying to debug why they are getting an error.


taiki-e (approved):
LGTM

PR #56, opened 2022-12-03
Future-proof the kevent API

asomers:
In FreeBSD 12, kevent grew some extra fields.  libc currently implements a FreeBSD 11 ABI, but that will change some day.  Tweak the kevent initialization so it will compile with either version.


notgull (approved):
Looks good to me! Once it's formatted, I'll merge it in.

taiki-e:
Thanks! Published in 2.5.2.

Issue #51, opened 2022-11-30
Potential bug when a source is closed while its poller is waiting


notgull:
Take this code:

```rust
use polling::{Poller, Event};
use std::net::TcpStream;
use std::thread::{sleep, spawn};
use std::time::Duration;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // Create a poller.
    let poller = Poller::new()?;

    // Register a TCP stream into it.
    let stream = TcpStream::connect(("google.com", 80))?;
    stream.set_nonblocking(true)?;
    let key = 1;
    poller.add(&stream, Event::readable(key))?;

    // Close the TCP stream on another thread.
    spawn(move || {
        sleep(Duration::from_secs(1));
        drop(stream);
    });

    // Wait for the stream to become readable.
    let mut events = Vec::new();
    poller.wait(&mut events, None)?;
    
    Ok(())
}
```

On Linux (`Linux 5.15.0-53-generic #59-Ubuntu`) this code simply hangs forever. I'm not sure how Windows reacts; I do not have a Windows machine accessible to me right now for testing, and Wine falls victim to a separate bug that I will be filing shortly (Edit: #52).

I realized that, once a file descriptor is registered into `polling`, there's no mechanism preventing that file descriptor from being deleted. While this is not an issue in most use cases (e.g. `async-io`), it may lead to buggy or even undefined behavior. The man page for `epoll_wait` says:

> For a discussion of what may happen if a file descriptor in an epoll instance being monitored by epoll_wait() is closed in another thread, see select(2).

The man page for `select` says (emphasis mine):

> If  a file descriptor being monitored by select() is closed in another thread, the result is unspecified. [...] On Linux  (and some other systems), closing the file descriptor in another thread has no effect on select(). **In summary, any application that relies on a particular behavior in this scenario must be considered buggy.**

I'm not entirely sure how `kqueue` or `wepoll` react in this case. In any case, I think that we need to decide:

* Should this behavior be allowed? This originally came up during the question of porting portions of this code to [`rustix`](https://crates.io/crates/rustix), which uses a system where the [`Epoll`](https://docs.rs/rustix/latest/rustix/io/epoll/struct.Epoll.html) (in this case) [either takes ownership of the source or borrows it for the duration of the polling process](https://docs.rs/rustix/latest/rustix/io/epoll/trait.Context.html).
[...]

PR #59, opened 2022-12-19
Add level and edge triggered modes to the poller

notgull:
At the moment, `polling` only supports oneshot-triggered event polling. However, most of the backends `polling` uses also supports level-triggered operation and some even support edge-triggered operations. While oneshot polling works for the `async-io` usecase, other users may want to use other polling modes. This PR adds alternative polling modes and avenues to use to react if one isn't available.

Note that we *may* be able to support level-triggered operation on Solarish, with some extra elbow grease. However, unless there's another way I couldn't see, the only way I could see to do this would be to keep track of the level-trigger file descriptors through a `HashSet` (wrapped in a `Mutex`) and that would add some performance burden to the polling operation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,8 +8,10 @@
 //! - [poll](https://en.wikipedia.org/wiki/Poll_(Unix)): VxWorks, Fuchsia, other Unix systems
 //! - [wepoll](https://github.com/piscisaureus/wepoll): Windows, Wine (version 7.13+)
 //!
-//! Polling is done in oneshot mode, which means interest in I/O events needs to be re-enabled
-//! after an event is delivered if we're interested in the next event of the same kind.
+//! By default, polling is done in oneshot mode, which means interest in I/O events needs to
+//! be re-enabled after an event is delivered if we're interested in the next event of the same
+//! kind. However, level and edge triggered modes are also available for certain operating
+//! systems. See the documentation of the [`PollMode`] type for more information.
 //!
 //! Only one thread can be waiting for I/O events at a time.
 //!
@@ -127,6 +129,40 @@ pub struct Event {
     pub writable: bool,
 }
 
+/// The mode in which the poller waits for I/O events.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[non_exhaustive]
+pub enum PollMode {
+    /// Poll in oneshot mode.
+    ///
+    /// In this mode, the poller will only deliver one event per file descriptor or socket.
+    /// Once an event has been delivered, interest in the event needs to be re-enabled
+    /// by calling `Poller::modify` or `Poller::add`.
+    ///
+    /// This is the default mode.
+    Oneshot,
+
+    /// Poll in level-triggered mode.
+    ///
+    /// Once an event has been delivered, polling will continue to deliver that event
+    /// until interest in the event is disabled by calling `Poller::modify` or `Poller::delete`.
+    ///
+    /// Not all operating system support this mode. Trying to register a file descriptor with
+    /// this mode in an unsupported operating system will raise an error. You can check if
+    /// the operating system supports this mode by calling `Poller::supports_level`.
+    Level,
+
+    /// Poll in edge-triggered mode.
+    ///
+    /// Once an event has been delivered, polling will not deliver that event again unless
+    /// a new event occurs.
+    ///
+    /// Not all operating system support this mode. Trying to register a file descriptor with
+    /// this mode in an unsupported operating system will raise an error. You can check if
+    /// the operating system supports this mode by calling `Poller::supports_edge`.
+    Edge,
+}
+
 impl Event {
     /// All kinds of events (readable and writable).
     ///
@@ -199,6 +235,16 @@ impl Poller {
         })
     }
 
+    /// Tell whether or not this `Poller` supports level-triggered polling.
+    pub fn supports_level(&self) -> bool {
+        self.poller.supports_level()
+    }
+
+    /// Tell whether or not this `Poller` supports edge-triggered polling.
+    pub fn supports_edge(&self) -> bool {
+        self.poller.supports_edge()
+    }
+
     /// Adds a file descriptor or socket to the poller.
     ///
     /// A file descriptor or socket is considered readable or writable when a read or write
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -125,6 +127,40 @@ pub struct Event {
...
+
+    /// Poll in edge-triggered mode.
+    ///
+    /// Once an event has been delivered, polling will not deliver that event again unless
+    /// a new event occurs.
+    ///
+    /// Not all operating system support this mode. Trying to register a file descriptor with
+    /// this mode in an unsupported operating system will raise an error. You can check if
+    /// the operating system supports this mode by calling `Poller::supports_edge`.
+    Edge,
```
zseri, line 161: wouldn't it make more sense to only provide these enum values if the platform supports them, instead of some weird fallback? And transform the `supports_level` and `supports_edge` "methods" into constants? (idk what an effective way to handle that in downstream build scripts or cfg stmts would be, or if downstream users simply could copy "our" `cfg` restrictions...

src/poll.rs
```diff
@@ -141,6 +153,19 @@ impl Poller {
...
+                        PollMode::Edge => {
+                            return Err(io::Error::new(
+                                #[cfg(not(polling_no_unsupported_error_kind))]
+                                io::ErrorKind::Unsupported,
+                                #[cfg(polling_no_unsupported_error_kind)]
+                                io::ErrorKind::Other,
+                                "edge-triggered events are not supported with poll()",
+                            ))
+                        }
+                    },
```
zseri, line 168: this match block is at least repeated once, and should be factored out.

src/poll.rs
```diff
@@ -141,6 +153,19 @@ impl Poller {
...
                     key: ev.key,
+                    remove: match mode {
+                        PollMode::Oneshot => true,
+                        PollMode::Level => false,
+                        PollMode::Edge => {
+                            return Err(io::Error::new(
+                                #[cfg(not(polling_no_unsupported_error_kind))]
+                                io::ErrorKind::Unsupported,
+                                #[cfg(polling_no_unsupported_error_kind)]
+                                io::ErrorKind::Other,
```
zseri, line 164: and this should be definitively factored out into something at the top-level (but obviously not publicly exported) to reuse it in all the backends which don't support some case.

zseri (changes requested):
It looks overall ok, but I dislike the code duplication in error cases.

zseri:
> the only way I could see to do this would be to keep track of the level-trigger file descriptors through a HashSet (wrapped in a Mutex)

this makes stuff way more complicated, because it introduces (additional) cases where the set of file descriptors can become out-of-sync with the OS view of file descriptors (e.g. a file descriptor is closed, and a new one is opened with the same ID), where the reactor would need to be notified; on other OSes this is handled mostly automatically by the OS view the native interface (although `epoll` on Linux has some cases regarding cloned file descriptors where this doesn't happen in the way usually expected); but it has a realistic chance of making it worse.

src/lib.rs
notgull: > wouldn't it make more sense to only provide these enum values if the platform supports them, instead of some weird fallback? 

It doesn't fall back, it errors out. This prevents cross-platform code from raising a compile-time error if, e.g, a certain code branch uses edge triggered mode.

The goal of this PR is to allow `polling` to be used in `calloop`, where edge triggered support already exists.

> And transform the `supports_level` and `supports_edge` "methods" into constants? 

The reason why I used methods instead of constants is so that is portable to potential future systems where edge/level triggered support is dependent on variable factors (e.g. kernel version).

notgull:
> this makes stuff way more complicated, because it introduces (additional) cases where the set of file descriptors can become out-of-sync with the OS view of file descriptors (e.g. a file descriptor is closed, and a new one is opened with the same ID), where the reactor would need to be notified; on other OSes this is handled mostly automatically by the OS view the native interface (although `epoll` on Linux has some cases regarding cloned file descriptors where this doesn't happen in the way usually expected); but it has a realistic chance of making it worse.

I see, thanks for confirming the gut feeling I had as to why we shouldn't do this. However, if this is an issue, isn't the `poll()` fallback backend also affected by this?

#51 is probably related.

src/lib.rs
```diff
@@ -569,3 +652,36 @@ cfg_if! {
...
+            target_os = "ios",
+            target_os = "tvos",
+            target_os = "watchos",
+            target_os = "freebsd",
+            target_os = "netbsd",
+            target_os = "openbsd",
+            target_os = "dragonfly",
+        ))
+    )
+))]
```
zseri, line 678: I don't think this overly large `cfg` restriction makes much sense, just put an `allow(unused)` here.

zseri (approved):
It looks now ok overall.

But could we add some tests to make sure that it works?

notgull:
It looks like there's an issue with `kqueue`, at least on MacOS. Let me spin up a FreeBSD VM so that I can check if it's there as well.

notgull:
Hmm, it looks like there's some subtlety with `EV_CLEAR` and changing modes in `kqueue` that I'm not aware of. On FreeBSD, for some reason, when I switch it to oneshot mode, it takes a while after the write to the TCP pipe for the new notification to be registered, so I have to add a non-zero timeout to the wait calls.

It's probably something obvious that I'm missing. I'll take another run at it in the morning.

tests/other_modes.rs
```diff
@@ -62,16 +64,17 @@ fn level_triggered() {
         .modify_with_mode(&reader, Event::readable(reader_token), PollMode::Oneshot)
         .unwrap();
 
-    writer.write_all(&data).unwrap();
+    writer.write(&data).unwrap();
     events.clear();
+    // BUG: Somehow, the notification here is delayed?
```
zseri, line 69: obvious place for a race condition, this might also be some context-switching/scheduling stuff...

did you try to look at how the (FreeBSD) kernel handles this internally?

zseri: how does our test differ from https://cgit.freebsd.org/src/tree/tests/sys/kqueue/libkqueue/read.c?h=stable/13#n175 ?

zseri:
uhm besides that... https://github.com/smol-rs/polling/blob/9a989d408b1c33d9fdb3b45d9631f65154df40be/src/kqueue.rs#L119-L127 looks really sus, especially since it can swallow events.

tests/other_modes.rs
zseri: does it make a practical/behavior difference for the timeout if `None` vs `Some(Duration::from_millis(0))` is specified?

zseri: I would be in favor of using `None` (= no timeout) here, or some really long timeout (e.g. 10 seconds). As this is the only problem remaining, but overall seems to have nothing to do with this PR as-is, I would also be in favor of merging this and creating a new issue for it.

notgull:
I think I've figured out the issue; sometimes, on `kqueue`, it takes a moment for the new events to be registered into queue after they're used. We just have to account for this in the tests.

At this stage, I think all of the issues with this PR are resolved.

PR #63, opened 2022-12-29
Test the backend that uses poll on CI

taiki-e:
This backend (https://github.com/smol-rs/polling/pull/26) is only used on minor operating systems which is not easy for us to test.

Historically we used to do these tests by adjusting cfg on Linux or macOS. (https://github.com/smol-rs/polling/pull/26#issue-770294269, https://github.com/smol-rs/polling/pull/26#pullrequestreview-741004153)

This patch makes testing in such a way easy and can be run on CI.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,7 +76,12 @@ macro_rules! syscall {
 }
 
 cfg_if! {
-    if #[cfg(any(target_os = "linux", target_os = "android"))] {
+    // Note: This cfg is intended to make it easy for polling developers to test
+    // the backend that uses poll, and is not a public API.
+    if #[cfg(polling_test_poll_backend)] {
+        mod poll;
+        use poll as sys;
+    } else if #[cfg(any(target_os = "linux", target_os = "android"))] {
         mod epoll;
         use epoll as sys;
     } else if #[cfg(any(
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -445,19 +450,22 @@ impl Poller {
...
-    target_os = "solaris",
-    target_os = "macos",
-    target_os = "ios",
-    target_os = "tvos",
-    target_os = "watchos",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd",
-    target_os = "dragonfly",
+#[cfg(all(
```
notgull, line 453: I'm getting mildly uncomfortable with the config sprawl happening here. The easiest way to resolve it would be to emit build constants in `build.rs`, but I'm not sure how portable that is.

notgull (approved):
Looks good to me!

src/lib.rs
taiki-e: As for cfg, another option is to move the implementation to the corresponding backend module, but doc-cfg does not work well either way.

- If we keep doc-cfg, this is the best I know of: https://github.com/smol-rs/polling/commit/c736dfb2aa054fc69e1bb8658a19acc71e1912cb
- If we remove doc-cfg, we can remove those cfgs: https://github.com/smol-rs/polling/commit/87def8d70a75a910f0fca223b0718a255bba5433

(I'm fine with either of the above.)

If we remove doc-cfg and use build-script, the code will be more little simpler (https://github.com/smol-rs/polling/commit/f3ee7469b001d281e1aa869e2240d61ae60a714a), but I don't much prefer it due to the compatibility issues mentioned in https://github.com/smol-rs/polling/pull/39#issuecomment-1237278819, some pitfalls specific to build scripts[^1], and the difficulty of finding typos in the build-script[^2].

[^1]: There are pitfalls specific to build scripts, such as parsing target triples directly instead of using `CARGO_CFG_*` makes it less robust. (In that case, it is mainly new or custom targets that are affected, though.)
[^2]: I have not yet introduced it in smol-rs's CI, but there is an unstable option to check for unrecognized cfgs called -Z check-cfg, and I use it in cfg-heavy projects like portable-atomic.

PR #70, opened 2023-01-06
test: Add test for multiple level triggered sources

notgull:


notgull:
Looks like it fails spuriously on MacOS.

PR #73, opened 2023-01-12
m: Use EVFILT_USER instead of a self-pipe on kqueue where supported

notgull:
Certain BSD platforms (macOS and FreeBSD) support an `EVFILT_USER` event on `kqueue`, which lets the user send arbitrary events to the queue to be dequeued. This PR adds a system that uses `EVFILT_USER` when available to notify the `Poller` instead of a self-pipe.

Draft because I need to test this on FreeBSD and NetBSD.


notgull:
Verified that this works on FreeBSD and NetBSD, although tests fail on NetBSD because it does not support high precision mode.

PR #74, opened 2023-01-12
m: Use port_send for event ports instead of a pipe

notgull:
Solarish's event ports support a [`poll_send`](https://www.unix.com/man-page/opensolaris/3C/port_send/) function for manually sending a notification to an event port. This PR uses that function to notify the event port, as opposed to a self-pipe.

Draft because I need to test this on an illumos box first.


notgull:
I've verified that tests pass on OpenIndiana.

PR #79, opened 2023-01-22
Test patched async-io on more platforms

taiki-e:
Split from #78 

~~Blocked on https://github.com/smol-rs/async-io/issues/110~~
Also the increase in time taken to test with vmactions is annoying.

after


notgull (approved):
Thanks!

Re: the vmactions taking forever. Maybe can only run them in specific circumstances? (Such as only if any of the `kqueue`-related files were affected).

taiki-e:
As for vmactions, I was thinking of trying [another action that may be faster](https://github.com/cross-platform-actions/action).

> Such as only if any of the `kqueue`-related files were affected

We have tried something similar in tokio (for loom-related files). However, the effect of this may be smaller, because these targets may also be affected by changes in tests and doc examples.

PR #72, opened 2023-01-09
feat: Expose kqueue's signal handling functionality

notgull:
This PR exposes functions that allow users to register signal handlers using `kqueue`.

See also: #68

PR #83, opened 2023-02-01
feat: Expose other kqueue filters

### LINKED ISSUE #68: Add support for custom kqueue filters
As suggested [here](https://github.com/Smithay/calloop/pull/123#issuecomment-1368543735), it may be beneficial to allow support for custom event filters for `kqueue`-targeting systems, such as `EVFILT_SIGNAL` for waiting on signals.

Maybe we could have an extension trait API, to the tune of:

```rust
pub struct Poller { /* ... */ }

pub mod os {
    #[cfg(uses_kqueue)]
    pub mod kqueue {
        pub trait PollerExt : Sealed {
            fn add_signal(/* ... */);
        }

        impl PollerExt for crate::Poller {
            /* ... */
        }
    }
}
```

notgull:
This PR uses extension traits to expose other `kqueue` filters, like signals and child processes.

Supercedes #72 

Closes #68


src/os.rs
```diff
@@ -0,0 +1,371 @@
...
+        target_os = "watchos",
+        target_os = "freebsd",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "dragonfly",
+    ),
+    not(polling_test_poll_backend),
+))]
+/// Functionality that is only available for `kqueue`-based platforms.
+pub mod kqueue {
```
zseri, line 17: why isn't this put into a separate file `src/os/kqueue.rs` ?

src/os.rs
```diff
@@ -0,0 +1,371 @@
...
+        ///
+        /// // Remove the filter.
+        /// poller.delete_filter(Signal::new(libc::SIGINT)).unwrap();
+        /// ```
+        fn delete_filter(&self, filter: impl Filter) -> io::Result<()>;
+    }
+
+    impl PollerSealed for Poller {}
+
+    impl PollerKqueueExt for Poller {
```
zseri, line 109: imo it would make more sense to tie this more directly to a filter, e.g. to the tuple `(Poller, F) where F: Filter` or `impl<F: Filter> PollerKqueueExt<F> for Poller`.

src/os/kqueue.rs
```diff
@@ -0,0 +1,329 @@
...
+        // Delete the filter.
+        self.poller.submit_changes([event])
+    }
+}
+
+/// A filter that can be registered into a `kqueue`.
+pub trait Filter: FilterSealed {}
+
+unsafe impl<T: FilterSealed + ?Sized> FilterSealed for &T {
+    fn filter(&self, flags: FilterFlags, key: usize) -> libc::kevent {
```
zseri, line 122: should be marked `#[inline(always)]`

src/os/kqueue.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// // Remove the filter.
+    /// poller.delete_filter(Signal::new(libc::SIGINT)).unwrap();
+    /// ```
+    fn delete_filter(&self, filter: F) -> io::Result<()>;
+}
+
+impl PollerSealed for Poller {}
+
+impl<F: Filter> PollerKqueueExt<F> for Poller {
+    fn add_filter(&self, filter: F, key: usize, mode: PollMode) -> io::Result<()> {
```
zseri, line 96: should be marked `#[inline(always)]`

src/os/kqueue.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+impl<T: Filter + ?Sized> Filter for &T {}
+
+/// Monitor this signal number.
+///
+/// No matter what `PollMode` is specified, this filter will always be
+/// oneshot-only.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Signal(c_int);
```
zseri, line 134: it would make more sense to just make the single wrapped field `pub` and omit the construct/destruct methods.

src/os/kqueue.rs
```diff
@@ -0,0 +1,329 @@
...
+/// Wait for a timeout to expire.
+///
+/// Modifying the timeout after it has been added to the poller will reset it.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Timer {
+    /// Identifier for the timer.
+    id: usize,
+
+    /// The timeout to wait for.
+    timeout: Duration,
```
zseri, line 227: consider marking these `pub`.

src/os/kqueue.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+impl Timer {
+    /// Create a new timer.
+    pub fn new(id: usize, timeout: Duration) -> Self {
```
zseri, line 232: if the fields of this struct would be marked `pub`, these methods could be omitted.

src/os/kqueue.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+impl Filter for Timer {}
+
+mod __private {
+    use crate::sys::FilterFlags;
+
+    #[doc(hidden)]
+    pub trait PollerSealed {}
```
zseri, line 320: it's probably a good idea to put this into `src/os.rs::__private`, as it should imo be used for every sealed poller extension. (and would make the pattern clear)

zseri (commented):
considerations regarding `#[inline]` and `pub`

zseri:
suggestion: also uhh it's suboptimal that you put 2 independent changes into the last commit (one which moved the file, and a fixup), that makes it harder to review, would be nice if you could avoid that in the future...

looks otherwise pretty ok now. thx

notgull:
Sorry about that, and thanks for the review!

Regarding the `pub` changes: I originally wanted to do this, but then I realize that would force us into a corner if new features are added to the `kqueue` API. For instance, if a new `fflags` option is added to a filter (this has happened relatively recently in NetBSD), we wouldn't be able to express it, since adding a new field to that structure would be a breaking change. Having the structs have private fields future-proofs us for future changes to `kqueue`.

zseri:
oh btw the reason I suggested `pub` is that if new features would be added, it would make sense that the API would need to adjusted anyways (although a builder-like pattern would be possible, it might make the current interface awkward).

notgull:
> if new features would be added, it would make sense that the API would need to adjusted anyways

Good point. I've updated it so the fields are `pub`, except for the `Process` filter, since that one is slightly more complex than the others.

zseri (approved):

Issue #82, opened 2023-02-01
Add support for anonymous pipes


notgull:
It would be useful if this crate supported Windows anonymous pipes.

I probably can't make any progress on this until #76 is complete, as this would require the backend to not be `wepoll`.

kchibisov:
Given that your patch merged, how far are we from anon pipes? Asking since rust time bombed mio 0.6.x (it has future incompatible code) and we must find an alternative.

notgull:
We still need to support direct `OVERLAPPED` operations in our use case. However, looking at your use case, it looks like you use [`mio-anonymous-pipes`], which "cheats" by putting the pipe onto a thread pool. Therefore, you might be able to get away with just wrapping the pipe in an [`Unblock`](https://docs.rs/blocking/latest/blocking/struct.Unblock.html) and using it like that.

notgull:
Not as much of a priority anymore as we've worked around this limitation in `alacritty`. This is mostly a duplicate of #97 now, as we would need overlapped operation support to truly support anonymous pipes in `polling`.

Issue #66, opened 2022-12-30
Expose the Events structure instead of using Vec<Event>


notgull:
My main complaint with the `polling` API is the fact that it doesn't expose `sys::Events`; instead, it reads events from the `sys::Events` into a `Vec<Event>`. In practice, this usually means that:

1. System events are read to `Events`.
2. `poller` events are read out of the `Events`.
3. `poller` events are appended to a `Vec<Event>`.
4. Downstream, the `poller` events are then [drained out of that `Vec<Event>`](https://github.com/smol-rs/async-io/blob/9b46fc0bfe9df6bc7f0f0d2dd0fe49122732c8cd/src/reactor.rs#L285-L312) and used to react to events.

Steps 1 and 2 are necessary, but steps 3 and 4 feel unnecessary. It would be nice if the `Events` structure was exposed and given a real API. Is there any particular reason why `polling` was designed to use `Vec<Event>` in the first place?

This would be a breaking change, but there's already breaking changes on the horizon (#38) so we should probably at least plan for this.

PR #88, opened 2023-02-16
m(windows): Reimplement Wepoll in Rust

### LINKED ISSUE #22: Removing the wepoll dependency?
Would it be possible to implement the windows backend directly in rust? I think making the code as accessible as possible to users is especially important with a library like this (mostly for debugging).

It looks like `mio` is written with `{nt, win}api`, and maybe it could be used as reference here.

### LINKED ISSUE #85: Null pointer dereference in patched wepoll code
See piscisaureus/wepoll#32. It looks like the patched wepoll code has a null pointer dereference for when the poller is notified.

notgull:
Thus far, `wepoll` has served us well in terms of polling on Windows. However, at this stage, it's time to migrate to our own home-rolled implementation. There is at least one critical soundness bug that stems from our usage of `wepoll` (#85), and it would be significantly easier to compile `polling` (as well as all of its dependents) if we purged our only C dependency. In addition, there are some issues that we can only solve if we move away from `wepoll` (smol-rs/async-io#17, smol-rs/async-io#25, #82).

The new strategy is similar to what `wepoll` currently does, except (hopefully) more robust. The core consists of using an I/O Completion Port as the core of the system, and then using the undocumented Auxillary Function Driver to deliver completion packets whenever a socket is considered "ready". There is a state object associated with each socket that contains the events currently being polled, the in-flight events associated with the socket, and whether the socket state is queued for deletion.

The differences between this implementation and `wepoll` includes:

- Rather than using a red-black tree to store socket state, we use a `HashMap`.
- A bounded `ConcurrentQueue` is used to hold pending socket updates, rather than a naive queue implementation. This queue has a capacity of 1024 elements and every update is run run the queue is full. In contrast, `wepoll`'s queue is unbounded.
- Packets have `Pin` guarantees, in order to prevent state captured by the I/O completion port from being moved.
- Instead of going through the `epoll_event` intermediary, we directly convert completion packets to `Event`s.
- We directly support notifying the `Poller` through a designated wakeup packet.
- We only expect one user to be polling at a time, so we use an `AtomicBool` instead of an atomic number.

Closes #22 and closes #85. Supersedes #76.


notgull:
I think this is stable enough for review now; the tests aren't catching any bugs that I can see (although the `precision` tests are taking a while? But that might just be part of the test).

taiki-e (commented):
Thanks! I have not reviewed the details regarding the Windows API, but at first glance, this looks great.

Also, we may want to test some downstream crates before merging/releasing this patch because such a rewrite may cause subtle differences in behavior. (in the similar [way that our CI tests async-io](https://github.com/smol-rs/polling/blob/d443196f641dfe83904dcc8bc8e163ec64180cdd/.github/workflows/ci.yml#L52-L59))

src/iocp/afd.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    unsafe fn unlock(self: Pin<&Self>) {
+        self.in_use.store(false, Ordering::SeqCst);
+    }
+}
+
+/// Get the base socket associated with a socket.
+pub(super) fn base_socket(sock: RawSocket) -> io::Result<RawSocket> {
+    // First, try the SIO_BASE_HANDLE ioctl.
+    let result = unsafe { try_socket_ioctl(sock, SIO_BASE_HANDLE) };
```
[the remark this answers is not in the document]
notgull: My bad, that was a poorly written safety comment. It's more like "the socket that's returned by the IOCTL must be valid." I changed it to be more clear.

notgull:
> Also, we may want to test some downstream crates before merging/releasing this patch because such a rewrite may cause subtle differences in behavior.

I just ran a small "crater test" on a few crates that have `polling` in their dependency tree. I ran `cargo test` for:

- `async-io`
- `async-net`
- `async-std`
- `isahc` 
- `libp2p` (had to figure out how to set up `protoc` on Windows for this 😮‍💨)
- `quinn` (tests mostly use `tokio` but the ones I modified to use `async-std` seemed to turn out OK)
- `zbus`

All of the tests seem to pass (or at least the ones that weren't already failing on Windows). I think we're fine in terms of potentially undefined behavior.

src/iocp/port.rs
```diff
@@ -0,0 +1,317 @@
...
+
+impl<T: CompletionHandle> OverlappedEntry<T> {
+    /// Convert into the completion packet.
+    pub(super) fn into_packet(self) -> T {
+        let packet = unsafe { self.packet() };
+        std::mem::forget(self);
+        packet
+    }
+
+    unsafe fn packet(&self) -> T {
```
fogti, line 275: what are the safety requirements of this function?

src/iocp/port.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    fn as_ptr(&self) -> *mut OVERLAPPED {
+        self.get_ref() as *const T as *mut OVERLAPPED
+    }
+}
+
+unsafe impl<T: Completion> CompletionHandle for Pin<Arc<T>> {
+    type Completion = T;
+
+    fn get(&self) -> Pin<&Self::Completion> {
```
fogti, line 84: why is the pointer that this function returns pinned?

src/iocp/afd.rs
```diff
@@ -0,0 +1,608 @@
...
+        } else {
+            let real_code = ntdll.RtlNtStatusToDosError(result);
+
+            Err(io::Error::from_raw_os_error(real_code as i32))
+        }
+    }
+}
+
+/// A one-time initialization cell.
+struct OnceCell<T> {
```
fogti, line 450: why do we roll our own `OnceCell` implementation here?

src/iocp/afd.rs
```diff
@@ -0,0 +1,608 @@
...
+        ptr::null_mut(),
+        0,
+        out.as_mut_ptr().cast(),
+        size_of::<RawSocket>() as u32,
+        bytes.as_mut_ptr(),
+        ptr::null_mut(),
+        None,
+    );
+
+    if result == SOCKET_ERROR {
```
fogti, line 603: is this really the only possible error?

src/iocp/afd.rs
```diff
@@ -0,0 +1,608 @@
...
+}
+
+/// Run an IOCTL on a socket and return a socket.
+///
+/// # Safety
+///
+/// The `ioctl` parameter must be a valid I/O control that returns a valid socket.
+unsafe fn try_socket_ioctl(sock: RawSocket, ioctl: u32) -> io::Result<RawSocket> {
+    let mut out = MaybeUninit::<RawSocket>::uninit();
+    let mut bytes = MaybeUninit::<u32>::uninit();
```
fogti, line 589: what do we gain by making this `u32` `MaybeUninit` here?

src/iocp/mod.rs
```diff
@@ -0,0 +1,820 @@
...
+        let deadline = timeout.and_then(|timeout| Instant::now().checked_add(timeout));
+        let mut packets = self.packets.lock().unwrap_or_else(|e| e.into_inner());
+        let mut notified = false;
+        events.packets.clear();
+
+        loop {
+            let mut new_events = 0;
+
+            // Indicate that we are now polling.
+            debug_assert!(!self.polling.swap(true, Ordering::SeqCst));
```
fogti, line 265: this won't be run during release builds, right?

src/iocp/mod.rs
```diff
@@ -0,0 +1,820 @@
...
+    pub(super) fn notify(&self) -> io::Result<()> {
+        // Push the notify packet into the IOCP.
+        self.port.post(0, 0, self.notifier.clone())
+    }
+
+    /// Run an update on a packet.
+    fn update_packet(&self, mut packet: Packet) -> io::Result<()> {
+        loop {
+            // If we are currently polling, we need to update the packet immediately.
+            if self.polling.load(Ordering::Acquire) {
```
fogti, line 324: so, will this then behave correctly on release builds, when `self.polling` never gets modified?

src/iocp/port.rs
notgull: It should only be called once, since it moved out the `T`, like `ManuallyDrop::take()`. I've clarified this in the docs.

src/iocp/port.rs
notgull: The pinning ensures that the overlapped object isn't moved while it's involved in an operation.

src/iocp/afd.rs
notgull: We can't use `once_cell` since its MSRV is 1.56, and pulling in `async_lock::OnceCell` would add a few too many dependencies for my liking. I figured rolling my own is the best option.

src/iocp/afd.rs
notgull: Yes; it's basically like returning `-1` for C functions.

src/iocp/afd.rs
notgull: Nothing, I guess. Fixed.

src/iocp/mod.rs
notgull: I'm not sure, I *think* the values are still consumed even on release mode? Either way, it's better to make it more explicit. Fixed!

src/iocp/port.rs
fogti: how could it be moved when it is behind an `&` reference?

src/iocp/afd.rs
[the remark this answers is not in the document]
fogti: huh, did the safety comment vanish?

src/iocp/port.rs
notgull: While it's being used in the overlapped operation, there is no reference (or at least, none that the borrow checker is aware of). As a safeguard against undefined behavior, this ensures that it isn't invalidated during the operation.

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+
+        loop {
+            let mut new_events = 0;
+
+            // Indicate that we are now polling.
+            let was_polling = self.polling.swap(true, Ordering::SeqCst);
+            debug_assert!(!was_polling);
+
+            let guard = CallOnDrop(|| {
+                debug_assert!(self.polling.swap(false, Ordering::SeqCst));
```
fogti, line 269: same here.

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+impl Events {
+    /// Creates an empty list of events.
+    pub(super) fn new() -> Events {
+        Events {
+            packets: Vec::with_capacity(1024),
+        }
+    }
+
+    /// Iterate over I/O events.
+    pub(super) fn iter(&self) -> impl Iterator<Item = Event> + '_ {
```
fogti, line 417: where is this used and why does it always clone packets?

src/iocp/afd.rs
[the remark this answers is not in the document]
notgull: It's here:

https://github.com/smol-rs/polling/blob/c3ab16109ce2a86aaafe1717c56deb3c83a5f9de/src/iocp/afd.rs#L586

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+                    if afd_data.handle_count() >= 1 {
+                        let events = afd_data.events();
+
+                        // If we closed the socket, remove it from being polled.
+                        if events.contains(AfdPollMask::LOCAL_CLOSE) {
+                            let source = {
+                                let mut sources =
+                                    poller.sources.write().unwrap_or_else(|e| e.into_inner());
+                                sources.remove(&socket_state.socket).unwrap()
+                            };
```
fogti, line 616: this nesting is a bit hard to read, spreading it out over multiple lines (including the currently implicit `drop`) might be clearer.

src/iocp/mod.rs
notgull: It's used in `lib.rs` for returning packets:

https://github.com/smol-rs/polling/blob/e85331c437eb5b9fce382e6a9a310e44a641a3d6/src/lib.rs#L500

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+        // Filter out events that the user didn't ask for.
+        {
+            if !socket_state.interest.readable {
+                event.readable = false;
+            }
+
+            if !socket_state.interest.writable {
+                event.writable = false;
+            }
+        }
```
fogti, line 638:
```suggestion
    event.readable &= socket_state.interest.readable;
    event.writable &= socket_state.interest.writable;
```
would imo be easier to comprehend

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+        Ok(return_value)
+    }
+
+    /// Begin deleting this socket.
+    fn begin_delete(self: Pin<Arc<Self>>) -> io::Result<()> {
+        // If we aren't already being deleted, start deleting.
+        let mut socket = self
+            .as_ref()
+            .socket_state()
+            .expect("can't delete notification packet");
```
fogti, line 668: this message could be confusing, as it is missing context that is also not obvious from the function name. A clear indication that the notification packet couldn't be deleted because it wasn't a socket might be a good idea.

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+
+            Arc::pin(IoStatusBlock::from(PacketInner::Socket {
+                packet: UnsafeCell::new(AfdPollInfo::default()),
+                socket: Mutex::new(state),
+            }))
+        };
+
+        // Keep track of the source in the poller.
+        {
+            let mut sources = self.sources.write().unwrap_or_else(|e| e.into_inner());
```
fogti, line 171: instead of doing such a weird dance around poisoning all the time in this code, wouldn't it be more appropriate to either pull in a library that already implements locking in such a way (e.g. `parking-lot`) or write an abstraction around `RwLock` (and maybe `Mutex`) once, that does this, and use it in all this code?

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+        // Get a reference to the source.
+        let source = {
+            let sources = self.sources.read().unwrap_or_else(|e| e.into_inner());
+
+            match sources.get(&socket) {
+                Some(s) => s.clone(),
+                None => {
+                    return Err(io::Error::from(io::ErrorKind::NotFound));
+                }
+            }
```
fogti, line 219: this might be cleaner done via chaining, e.g.
```suggestion
            sources.get(&socket)
                .as_ref()
                .cloned()
                .ok_or_else(|| io::Error::from(io::ErrorKind::NotFound))?
```

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+        log::trace!("remove: handle={:?}, sock={}", self.port, socket);
+
+        // Get a reference to the source.
+        let source = {
+            let mut sources = self.sources.write().unwrap_or_else(|e| e.into_inner());
+
+            match sources.remove(&socket) {
+                Some(s) => s,
+                None => {
+                    // Just return.
```
fogti, line 241: this comment is unhelpful

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+        };
+
+        // Only drain the queue's capacity, since this could in theory run forever.
+        for _ in 0..max {
+            if let Ok(packet) = self.pending_updates.pop() {
+                packet.update()?;
+            } else {
+                return Ok(());
+            }
+        }
```
fogti, line 356:
```suggestion
        core::iter::from_fn(|| self.pending_updates.pop().ok())
            .take(max)
            .map(|packet| packet.update())
            .collect::<Result<Vec<()>,_>>()?;
```
the `Vec<()>` shouldn't actually allocate anything, as `()` is ZST.

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+
+        // Create a new AFD instance.
+        let afd = Arc::new(Afd::new()?);
+
+        // Register the AFD instance with the I/O completion port.
+        self.port.register(&*afd, true)?;
+
+        // Insert a copy of the AFD instance into the list.
+        afd_handles.push(afd.clone());
```
fogti, line 381: do we ever prune AFDs from `afd_handles`? could this exhaust all available memory? There currently doesn't seem to be a way to tell the poller that it should `shrink_to_fit`, free all cached blocks that aren't used anymore (or, if I understand the usage of `afd_handles` correctly, it seems to act as a cache and we could get away with
```rust
self.afd = Mutex::new(Vec::new());
```

src/iocp/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+pub(super) struct Events {
+    /// List of IOCP packets.
+    packets: Vec<Event>,
+}
+
+unsafe impl Send for Events {}
+
+impl Events {
+    /// Creates an empty list of events.
+    pub(super) fn new() -> Events {
```
fogti, line 410: why isn't this `pub`?

src/iocp/mod.rs
fogti: [`debug_assert!`](https://doc.rust-lang.org/core/macro.debug_assert.html) gets completely eliminated on release builds. *Beware*.

src/iocp/mod.rs
fogti: hmm, always cloning the packets before the filter does seem contra-productive.

src/iocp/mod.rs
notgull: Each AFD handle can be associated with up to 32 sockets. We use the ref-counting mechanism in `Arc` to keep track of how many sockets are being used. This ensures that old AFDs are reused by old sockets. I guess you could end up with thousands of AFD handles if you open and close 100,000's of sockets. In that case, I can add a pruning mechanism for when there are a large number of AFD handles with no associated sockets.

src/iocp/mod.rs
notgull: `Events` is internal to the `Poller` struct and used as an event container. That being said, I'd like to make the `Events` struct public so that we don't go through a `Vec<Event>` for events, see #66.

src/iocp/mod.rs
notgull: I think so too! This could be avoided if the `Events` struct was `pub` and could be read from directly, but this would be a breaking change. See #66.

src/iocp/mod.rs
fogti: alternatively, you could "just" use weak pointers, that way, opening, polling, and then closing 100'000's of sockets should automatically free the associated AFDs, and the next crawl of the list could then prune the gone ones.

src/iocp/mod.rs
notgull: Hmm, yeah, it does look pretty ugly. I've added a `lock!` macro in order to do it automatically.

src/iocp/mod.rs
notgull: I've replaced the block with a more concise version.

src/iocp/mod.rs
fogti: ok, so `Event` can actually be copied (so I think that should be used instead), then it doesn't really matter much (except that, of course, the mixing of `Events` and `Vec<Event>` instead of one just being a `type` alias of the other or such is unfortunate)

src/iocp/mod.rs
notgull: I changed it to clarify what it's doing, making it more helpful

src/iocp/mod.rs
```diff
@@ -45,20 +45,30 @@ use std::marker::PhantomPinned;
...
+use std::sync::{Arc, Mutex, MutexGuard, RwLock, Weak};
 use std::time::{Duration, Instant};
 
 #[cfg(not(polling_no_io_safety))]
 use std::os::windows::io::{AsHandle, BorrowedHandle};
 
+/// Macro to lock and ignore lock poisoning.
+macro_rules! lock {
+    ($lock_result:expr) => {{
+        ($lock_result).unwrap_or_else(|e| e.into_inner())
```
fogti, line 57: Rust has hygienic macros, so the parentheses around `$lock_result` are unnecessary.

src/iocp/mod.rs
notgull: Great idea! I used `try_for_each` instead of `.map.collect` since it's more idiomatic.

Unrelated: Should we implement an `Iterator` for draining a `ConcurrentQueue` in order to avoid using `from_fn()`?

src/iocp/port.rs
fogti: hm, seems a bit brittle, but I'll go with it for now.

src/iocp/mod.rs
fogti: would be ok, it should be documented that the `Iterator::next`, analogous to the rest of the interface, won't block.

fogti:
Thanks for the work btw. Looks impressive.

fogti (approved):

notgull:
> Thanks for the work btw. Looks impressive.

Thanks for the review! I know this is a monster PR, so I appreciate you taking the time to sit down and look through it all. I'll merge once the CI passes.

fogti:
It might be a good idea to do another downstream test, otherwise this looks ok. I think it is appropriate to squash-merge this when done. The possible brittleness of `<Pin<Arc<T>> as CompletionHandle>::get()` (or maybe the whole `CompletionHandle` + `OVERLAPPED` API) should imo be mentioned/appended to the intro post in the resulting merge commit.
"run run" in the intro post looks like a mistake btw.

> We only expect one user to be polling at a time, so we use an AtomicBool instead of an atomic number.

Can/do we enforce that?

notgull:
> > We only expect one user to be polling at a time, so we use an AtomicBool instead of an atomic number.
> 
> Can/do we enforce that?

A `Mutex` in the top-level `Poller` is locked before any polling is done.

https://github.com/smol-rs/polling/blob/e85331c437eb5b9fce382e6a9a310e44a641a3d6/src/lib.rs#L491-L505

PR #96, opened 2023-03-13
feat: Add edge/oneshot combination mode

### LINKED ISSUE #94: Allow mode combining when asking for polling with mode.
With at least epoll, `EPOLLONESHOT` could be used on its own(meaning level triggered) or with `Edge`. Right though, you can only do oneshot level triggering, because it simply passed plain oneshot.

notgull:
Closes #94 by adding a new `PollMode` that combines edge triggered and oneshot-triggered mode, available on edge-triggered-enabled platforms. This isn't a breaking change, since `PollMode` is `non_exhaustive`.

### DIFF
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -222,6 +222,7 @@ impl Poller {
                 PollMode::Oneshot => libc::EPOLLONESHOT,
                 PollMode::Level => 0,
                 PollMode::Edge => libc::EPOLLET,
+                PollMode::EdgeOneshot => libc::EPOLLONESHOT | libc::EPOLLET,
             };
             if ev.readable {
                 flags |= read_flags();
--- a/src/iocp/mod.rs
+++ b/src/iocp/mod.rs
@@ -146,7 +146,7 @@ impl Poller {
         );
 
         // We don't support edge-triggered events.
-        if matches!(mode, PollMode::Edge) {
+        if matches!(mode, PollMode::Edge | PollMode::EdgeOneshot) {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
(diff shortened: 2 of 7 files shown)


notgull:
The Android build is failing; is this another regression from 1.68?

taiki-e:
> The Android build is failing; is this another regression from 1.68?

Yeah, it's https://github.com/cross-rs/cross/issues/1128 / https://github.com/rust-lang/rust/issues/103673.
It can be worked around by [using unreleased images of cross](https://github.com/taiki-e/portable-atomic/blob/b91c04ca3d25144ce3f6b4bb7043ffd856167c7c/Cross.toml#L20-L22).

notgull:
Closing and reopening to re-run CI after smol-rs/polling#99

taiki-e (approved):

Thomasdezeeuw:
I'm coming here from Alacritty and I just wanted to say that Mio had its shared of issues using oneshot:
* https://github.com/tokio-rs/mio/issues/911
* https://github.com/tokio-rs/mio/issues/808
* https://github.com/tokio-rs/mio/issues/394
We ended up removing support altogether: https://github.com/tokio-rs/mio/issues/928. Just an FYI.

notgull:
Hello!

Oneshot mode was actually the only mode that `polling` used at first; other modes were only added recently, see #59. This is because its largest consumer, the `async-io` crate, is built on oneshot mode, as opposed to `tokio`'s reactor using edge-triggered mode.

Thanks for telling us about these issues; I will write some tests to make sure that we don't suffer from them. I don't *think* we have any issues with tokio-rs/mio#394, but I'd have to check.

PR #99, opened 2023-03-18
ci: Fix Android breakage

notgull:
Uses a workaround for cross-rs/cross#1128 and rust-lang/rust#103673


taiki-e:
We maybe also need to update cross version to the latest.

From:

https://github.com/smol-rs/polling/blob/e10c7e8da1aade91fa27cb953b34462dbb6a24b9/.github/workflows/ci.yml#L79-L82

To:

```yaml
       uses: taiki-e/install-action@cross
```

notgull:
Hmm, looks like the new version broke the other targets. Not sure if there's a way around this.

taiki-e:
I think `rustup target add`+`cargo build --target` would work for targets that cross does not support. (This crate is a library and does not require a linker or libraries to build the library part.)

https://github.com/smol-rs/polling/blob/e10c7e8da1aade91fa27cb953b34462dbb6a24b9/.github/workflows/ci.yml#L100-L101

notgull:
Looks like the FreeBSD/NetBSD builds are failing here... actually, can we just remove them, since we're testing them anyways using Cirrus/vmactions?

taiki-e:
Yeah, there is no need to build NetBSD and FreeBSD by using cross as they have been tested elsewhere.

https://github.com/smol-rs/polling/blob/e10c7e8da1aade91fa27cb953b34462dbb6a24b9/.github/workflows/ci.yml#L103
https://github.com/smol-rs/polling/blob/e10c7e8da1aade91fa27cb953b34462dbb6a24b9/.cirrus.yml#L13

taiki-e:
> Yeah, there is no need to build NetBSD and FreeBSD by using cross as they have been tested elsewhere.

It may be a good idea to leave a comment on cross job about we are testing BSDs using Cirrus/vmactions.
https://github.com/smol-rs/polling/blob/e10c7e8da1aade91fa27cb953b34462dbb6a24b9/.github/workflows/ci.yml#L68


---

Also you may need to change the target triple for Fuchsia from x86_64-fuchsia to x86_64-unknown-fuchsia.

https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1680-2023-03-09

> Add vendor to Fuchsia's target triple

taiki-e (approved):
Thanks!

PR #101, opened 2023-03-29
feat: Add functionality for posting events to the IOCP

notgull:
Adds a `CompletionPacket` structure that can be used to post arbitrary events to the I/O completion port. The idea is to allow for self-wakeups without using pipes, which aren't available on Windows.

### DIFF
--- a/src/iocp/mod.rs
+++ b/src/iocp/mod.rs
@@ -458,11 +458,11 @@ impl CompletionPacket {
     }
 
     /// Get the event associated with this packet.
-    pub fn event(&self) -> Event {
+    pub fn event(&self) -> &Event {
         let data = self.0.as_ref().data().project_ref();
 
         match data {
-            PacketInnerProj::Custom { event } => *event,
+            PacketInnerProj::Custom { event } => event,
             _ => unreachable!(),
         }
     }
--- a/src/iocp/mod.rs
+++ b/src/iocp/mod.rs
@@ -324,8 +324,8 @@ impl Poller {
     }
 
     /// Push an IOCP packet into the queue.
-    pub(super) fn post(&self, packet: &CompletionPacket) -> io::Result<()> {
-        self.port.post(0, 0, packet.clone().0)
+    pub(super) fn post(&self, packet: CompletionPacket) -> io::Result<()> {
+        self.port.post(0, 0, packet.0)
     }
 
     /// Run an update on a packet.
--- a/src/iocp/mod.rs
+++ b/src/iocp/mod.rs
@@ -323,6 +323,11 @@ impl Poller {
         self.port.post(0, 0, self.notifier.clone())
     }
 
+    /// Push an IOCP packet into the queue.
+    pub(super) fn post(&self, packet: &CompletionPacket) -> io::Result<()> {
+        self.port.post(0, 0, packet.clone().0)
+    }
+
     /// Run an update on a packet.
     fn update_packet(&self, mut packet: Packet) -> io::Result<()> {
         loop {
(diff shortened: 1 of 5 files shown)


src/iocp/mod.rs
```diff
@@ -323,6 +323,11 @@ impl Poller {
         self.port.post(0, 0, self.notifier.clone())
     }
 
+    /// Push an IOCP packet into the queue.
+    pub(super) fn post(&self, packet: &CompletionPacket) -> io::Result<()> {
+        self.port.post(0, 0, packet.clone().0)
+    }
```
fogti, line 329: why do we clone the packet (and don't pass it as is)? It doesn't have a special `Clone` or `Drop` implementation, so this shouldn't be necessary.

src/iocp/mod.rs
```diff
@@ -442,6 +447,27 @@ impl Events {
...
+    pub fn new(event: Event) -> Self {
+        Self(Arc::pin(IoStatusBlock::from(PacketInner::Custom { event })))
+    }
+
+    /// Get the event associated with this packet.
+    pub fn event(&self) -> Event {
+        let data = self.0.as_ref().data().project_ref();
+
+        match data {
+            PacketInnerProj::Custom { event } => *event,
```
fogti, line 465: why do we copy the event here and don't return a reference?

notgull: Because `Event` is a `Copy` struct... although it might not be in the future. Fixed.

src/iocp/mod.rs
notgull: `Packet` is a typedef for `Pin<Arc<...>>`, so that the underlying packet internals aren't dropped while the packet is waiting in the AFD system for an event. `IoCompletionPort::post()` expects this `Pin<Arc<...>>` for this reason.

fogti: how is this expected to be used?

notgull: The general idea that it can be used for events that are out of band of the `Poller`, for instance:

```rust
let poller = Arc::new(Poller::new().unwrap());
let my_source: TcpStream = /* ... */;
#[cfg(unix)]
let external_source: EventFd = /* ... */;
let my_key: usize = 0;
let external_key: usize = 1;

// Register the sources.
poller.add(my_source, Event::readable(my_key));
#[cfg(unix)]
poller.add(external_source, Event::readable(external_key));

// Wait for something to happen.
thread::spawn({
    let poller = poller.clone();
    move || {
        #[cfg(windows)]
        let packet = CompletionPacket::new(Event::readable(external_key));

        wait_for_something_blocking_to_happen();
        #[cfg(unix)]
        // On unix, we can use pipes/eventfds, like so:
        external_source.signal();

        #[cfg(windows)]
        // On windows, we don't have an equivalent to eventfd/pipe, so we need to
        // post the event
        poller.post(&packet);
    }
});

loop {
    let mut events = vec![];
    poller.wait(&mut events, None)?;

    for event in events {
        if event.key == my_key {
            handle_my_source(&my_source)?;
        } else if event.key == external_key {
            handle_external_source()?;
        }
    }
}
```

The motivating use case is to be able to shove Windows anonymous pipes into another thread but still be able to wait on them. See [here](https://github.com/alacritty/alacritty/pull/6846#issuecomment-1500933116) for more info.

notgull: Realized that you might be asking why I took this API design pattern instead of just taking an `Event`.

Imagine that, instead of the above `thread::spawn()` block being a one-off wait, that it's instead a loop that calls `wait_for_something_blocking_to_happen()`, then calling the `post()` function. If the `post()` function just took an `Event`, it would have to re-allocate the `Packet` every time it posted an event. With this strategy, no extra allocation has to occur.

fogti:
ok idk I probably can't accurately test this; I have no direct access to any windows machine, and even if I could I currently don't have time to bother to freshly setup rust and get stuff to work there...

code-wise, I count this as "should be good enough". LGTM. might just lack in testing.

tests/windows_post.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    thread::spawn({
+        let poller = Arc::clone(&poller);
+        move || {
+            for i in 0..3 {
+                poller
+                    .post(CompletionPacket::new(Event::writable(i)))
+                    .unwrap();
+
+                thread::sleep(Duration::from_millis(100));
```
fogti, line 39: the timing of stuff here looks odd...

tests/windows_post.rs
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+    });
+
+    for i in 0..3 {
+        poller
+            .wait(&mut events, Some(Duration::from_secs(5)))
+            .unwrap();
+
+        assert_eq!(events.len(), 1);
```
fogti, line 49: could we instead resort to just `pop`ing a single element in each iteration to make this more robust?

related: why don't events get coalesced/bursted here?

notgull: > could we instead resort to just `pop`ing a single element in each iteration to make this more robust?

Sounds good!

> related: why don't events get coalesced/bursted here?

The timeout above stops them from being coalesced.

fogti (approved):
LGTM. (failed tests are network timeouts on githubs intranet)

PR #106, opened 2023-04-10
Use Dependabot to update Github Actions

notgull:
Adds configuration for Dependabot to also propose updates to our Github Actions setup.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,10 @@ updates:
     commit-message:
       prefix: ''
     labels: []
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    commit-message:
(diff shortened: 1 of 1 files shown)


taiki-e:
FWIW, I previously used Dependabot for Github Actions in my project but disabled it because it was buggy: https://github.com/taiki-e/install-action/issues/43#issuecomment-1372291880

That bug may have been fixed in https://github.com/dependabot/dependabot-core/pull/6743, but I have not confirmed it yet.

taiki-e:
I have confirmed that the bug I was concerned about has been fixed (https://github.com/taiki-e/install-action/issues/43#issuecomment-1927035987), so we may want to revisit this.

(Hopefully, my fight with the dependabot bugs is over...)
dependabot

PR #108, opened 2023-04-18
Replace libc with rustix in some backends

notgull:
Replaces `libc` with `rustix` in the `epoll`, event ports and `poll`-based backends. This ensures that `libc` is entirely bypassed when making system calls.

We can't do `kqueue` until bytecodealliance/rustix#605 is released.

Supersedes #91


notgull:
...and since bytecodealliance/rustix#605 is released now, `kqueue` can be implemented with `rustix` now too! (Or almost all of it). Thanks, @sunfishcode!

Cargo.toml
```diff
@@ -30,6 +30,7 @@ autocfg = "1"
 
 [target.'cfg(any(unix, target_os = "fuchsia", target_os = "vxworks"))'.dependencies]
 libc = "0.2.77"
```
taiki-e, line 32: Can we move this to dev-deps?

notgull: It's still used by the kqueue backend. I can make it kqueue only if you want.

taiki-e (approved):
LGTM

PR #109, opened 2023-04-18
Add level triggered example

notgull:
Adds an example of how to use the `Poller` in level triggered mode.

WIP


notgull:
Needs a rework

PR #110, opened 2023-04-28
Use Cirrus CI for NetBSD/OpenBSD

taiki-e:
Follow-up to https://github.com/smol-rs/polling/pull/79.

This also drops slow/flaky vmactions.


taiki-e:
Before:
before

After:
after

Aboud 4x faster.

Issue #114, opened 2023-05-31
[examples] two-listeners; add message


Horki:
it would be cool if we had a message on connecting to the server like in https://github.com/tokio-rs/mio/blob/master/examples/udp_server.rs#L32-L34

notgull:
Closed by #115

PR #115, opened 2023-05-31
114: Examples; add nc message in 2 listeners

Horki:
#114


notgull (approved):
Thanks!

The wine failure looks spurious to me.

PR #117, opened 2023-06-11
Bump MSRV to 1.63

notgull:
cc smol-rs/smol#244

Removes the build script and all of the conditional features


notgull:
The wine failure is taiki-e/setup-cross-toolchain-action#9

taiki-e (approved):

Issue #38, opened 2022-08-11
Use `AsFd` and `AsSocket` instead of `AsRawFd` and `AsRawSocket`


notgull:
Rust 1.63 stabilized the I/O safety features in the standard library. As per sunfishcode/io-lifetimes#38, Rust crates in the ecosystem (such as `async-std` and `smol`) should be migrated to using the new I/O-safe traits instead of the `AsRaw*` family of traits.

To port this crate, we could replace the `Source` trait implementations, currently:

```rust
        impl Source for RawFd {
            fn raw(&self) -> RawFd {
                *self
            }
        }

        impl<T: AsRawFd> Source for &T {
            fn raw(&self) -> RawFd {
                self.as_raw_fd()
            }
        }
```

with this:

```rust
        impl Source for BorrowedFd<'_> {
            fn raw(&self) -> RawFd {
                self.as_raw_fd()
            }
        }

        impl<T: AsFd> Source for &T {
            fn raw(&self) -> RawFd {
                self.as_fd().as_raw_fd()
            }
        }
```

Although, ideally, the trait would return `BorrowedFd<'_>` instead, and the rest of the crate would be adjusted accordingly.

I am willing to implement this change in this crate as well as in `async-io`. The main question is if the MSRV bump is worth it.

taiki-e:
If we do this, it would be a breaking change because it would change types that implement `Source `. And the new version force downstream to use `AsFd`/`AsSocket`. Right?

notgull:
I agree, this might be premature. Let's wait until the ecosystem evolves to act.

notgull:
Okay, the ecosystem has evolved. We can now definitely use `AsFd`/`AsSocket` in our definitions. However, it comes with a catch; for `kqueue`, IOCP, `poll` and event ports, the source (the `AsFd` implementer) has to outlive its registration in the `Poller`. I'm still brainstorming the best way of implementing this in a way that is compatible with both `async-io` and `calloop`.

notgull:
Unfortunately I can't escape the conclusion that integrating I/O safety into this API would require it becoming either unsafe or very unwieldy. For any given `AsFd` implementor, there is no guarantee that the file descriptor it returns will be consistent or that it won't be closed before the actual object is dropped. The other way of doing this would be to parameterize the API around `BorrowedFd`, but that would make the API hard to use, especially for `async-io` and `calloop`.

notgull:
[...]

PR #123, opened 2023-06-25
Rework the API for I/O safety

notgull:
This reworks the `polling` API with I/O safety in mind. See #38 and smol-rs/smol#244 for more information.

This is a breaking change that makes the `add()` family of functions `unsafe`. I am very open to alternatives to this change.

The indirect effects of this change is that is makes `async_io::Async::get_mut` and `read/write_with_mut` unsafe, as well as having implications for the `AsyncRead/AsyncWrite` impls as well. See smol-rs/async-io#142. I would appreciate community feedback on these changes.

### DIFF
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -5,8 +5,9 @@ use std::io;
 use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd};
 use std::time::Duration;
 
+use rustix::event::{epoll, eventfd, EventfdFlags};
 use rustix::fd::OwnedFd;
-use rustix::io::{epoll, eventfd, read, write, EventfdFlags};
+use rustix::io::{read, write};
 use rustix::time::{
     timerfd_create, timerfd_settime, Itimerspec, TimerfdClockId, TimerfdFlags, TimerfdTimerFlags,
     Timespec,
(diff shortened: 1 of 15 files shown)


src/epoll.rs
```diff
@@ -105,7 +111,7 @@ impl Poller {
     }
 
     /// Modifies an existing file descriptor.
-    pub fn modify(&self, fd: RawFd, ev: Event, mode: PollMode) -> io::Result<()> {
+    pub fn modify(&self, fd: BorrowedFd<'_>, ev: Event, mode: PollMode) -> io::Result<()> {
```
sunfishcode, line 114: If `add` is `unsafe` and takes a `RawFd`, then `modifiy` should be too, for the same reason.

src/epoll.rs
```diff
@@ -85,7 +87,11 @@ impl Poller {
     }
 
     /// Adds a new file descriptor.
-    pub fn add(&self, fd: RawFd, ev: Event, mode: PollMode) -> io::Result<()> {
+    ///
+    /// # Safety
+    ///
+    /// The `fd` must be a valid file descriptor.
```
sunfishcode, line 93: It'd be good to also mention that the file descriptor must remain valid as long as it's in the `Poller`.

src/epoll.rs
sunfishcode: Ah, good point. I was somehow thinking it could modify the stored file descriptor, but as you say, it can't.

src/epoll.rs
notgull: This isn't true for `epoll` only. I've added documentation clarifying this.

notgull:
@smol-rs/admins Can anyone review this PR? Given that this change will have ecosystem wide effects, I'd like to make sure this is what we want.

fogti:
Make sure to also review the corresponding PR for `async-io`. I think most of it looks good. idk if/what it would do to code further downstream to `async-io` and others. (API changes like the new `IoSafe` trait should maybe documented better /+ more prominently.

notgull:
@taiki-e You have good judgement when it comes to this sort of thing, so I'll defer to you before I merge this.

notgull:
I'll merge this now and we can discuss the changes in greater depth later.

PR #124, opened 2023-06-29
Remove the bitflags dependency

notgull:
The `bitflags` crate is planning on doing a breaking change, see bitflags/bitflags#369. This is a little bit too much version churn for my liking, and it's not like we expose it publicly. This PR removes it from the Windows backend.


taiki-e (approved):

KodrAus:
Yep, it’s been a bit of a churning journey trying to tighten things up. Avoiding dependencies though is a good thing, and it’s nice you were able to replace it with almost the same amount of code!

PR #128, opened 2023-08-03
Support for the ESP-IDF framework

ivmarkov:
(This has been long supported in a [fork](https://github.com/smol-rs/polling/compare/master...esp-rs-compat:polling:master) of the `polling` crate, but the fork was against an ancient version. So I finally jumped the gun to rebase it against the latest `polling` and try to upstream now.)

ESP IDF already does support the `poll` syscall, so this PR is only necessary because - ESP IDF being an embedded OS (or toolkit?) - does not support anything process-related, including the `pipe` syscall.
 
However [it does support a subset of the `eventfd` syscall API](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/vfs.html#event-fds), so what this PR does is to augment the `poll` module - specifically for ESP IDF - with an alternative `eventfd` codepath to where `pipe` is usually utilized. The same idea (use `eventfd` for the ESP IDF) was I believe re-used by @jasta in the [recently upstreamed support for ESP IDF in tokio's reactor](https://github.com/tokio-rs/mio/pull/1692/files#diff-f9d8cbc0b569bc0ec06f40054daee6950e838fba27849dffefc3c74803118223L89).

In terms of changes, two main items:
* As per above, the `poll` module is augmented with alternative code-path based on `eventfd`
* I had to also update the `rustix` dependency from `0.37.X` to `0.38.X` as the support for ESP IDF in `rustix` (including the `eventfd` syscall) lands in the `0.38.X` version. I hope you don't mind and I noticed other crates in the `smol` ecosystem already use `rustix` `0.38.X` so I guess it was just a matter of time and somebody taking the effort to update it for this crate as well
  * The changes for the new `rustix` version are really mainly addressing latest `rustix` shuffling code around (the new `event` module). Two changes worth noting are [this one](https://github.com/smol-rs/polling/pull/128/files#diff-1aed9be8eb5b536b6f0da7ee8d523f7415dc55c02ba340c5dae388e891dfc75fR121) and [this one](https://github.com/smol-rs/polling/pull/128/files#diff-1aed9be8eb5b536b6f0da7ee8d523f7415dc55c02ba340c5dae388e891dfc75fR162) where I believe in the latter, the cast to `time_t` was incorrect in the first place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,12 @@ jobs:
           RUSTFLAGS: ${{ env.RUSTFLAGS }} --cfg polling_test_poll_backend
         if: startsWith(matrix.os, 'ubuntu')
       - run: cargo hack build --feature-powerset --no-dev-deps
+      - name: Add rust-src
+        if: matrix.rust == 'nightly'
+        run: rustup component add rust-src
+      - name: Check selected Tier 3 targets
+        if: matrix.rust == 'nightly' && matrix.os == 'ubuntu-latest'
+        run: cargo +nightly check -Z build-std --target=riscv32imc-esp-espidf --config 'patch.crates-io.rustix.git="https://github.com/bytecodealliance/rustix"'
 
   cross:
     runs-on: ${{ matrix.os }}
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -103,7 +103,7 @@ impl Poller {
 
         let span = tracing::trace_span!(
             "add",
-            notify = ?self.notify,
+            notify_read = ?self.notify.fd().as_raw_fd(),
             ?fd,
             ?ev,
         );
@@ -142,7 +142,7 @@ impl Poller {
 
         let span = tracing::trace_span!(
             "modify",
-            notify_read = ?self.notify,
+            notify_read = ?self.notify.fd().as_raw_fd(),
             ?fd,
             ?ev,
         );
@@ -176,7 +176,7 @@ impl Poller {
 
         let span = tracing::trace_span!(
             "delete",
-            notify = ?self.notify,
+            notify_read = ?self.notify.fd().as_raw_fd(),
             ?fd,
         );
         let _enter = span.enter();
@@ -202,7 +202,7 @@ impl Poller {
     pub fn wait(&self, events: &mut Events, timeout: Option<Duration>) -> io::Result<()> {
         let span = tracing::trace_span!(
             "wait",
-            notify = ?self.notify,
+            notify_read = ?self.notify.fd().as_raw_fd(),
             ?timeout,
         );
         let _enter = span.enter();
@@ -298,7 +298,7 @@ impl Poller {
     pub fn notify(&self) -> io::Result<()> {
         let span = tracing::trace_span!(
             "notify",
-            notify = ?self.notify,
+            notify_read = ?self.notify.fd().as_raw_fd(),
         );
         let _enter = span.enter();
 
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -74,7 +74,7 @@ impl Poller {
                 poll_fds: vec![PollFd::from_borrowed_fd(
                     // SAFETY: `notify.fd()` will remain valid until we drop `self`.
                     unsafe { BorrowedFd::borrow_raw(notify.fd().as_raw_fd()) },
-                    PollFlags::RDNORM,
+                    notify.poll_flags(),
                 )],
                 fd_data: HashMap::new(),
             }),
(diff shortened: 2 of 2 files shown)


notgull (changes requested):
I don't like the way that the alternating `pipe`/`eventfd` code is scattered throughout the module. It would be better if there was a single `Notifier` type that contained either the pipe on non-ESP and the eventfd on ESP. I did a similar thing in `kqueue.rs` to switch between pipes and user events.

Looks good otherwise.

ivmarkov:
OK. Will schedule some time over the weekend to do that.

daggerrz:
@ivmarkov, this is a related: https://github.com/smol-rs/polling/pull/129

I rant into this while using your branch for embedded development and building tests for my MacOS host. The Rustix dependency update breaks some targets.

notgull:
I merged #123 because I don't want to build up changes that are going to get stale. Just a heads up. This also takes care of the rustix v0.38 port.

daggerrz:
Ah, thanks, did not see that it was already pending!

ivmarkov:
@notgull Two additional follow up questions:
* Looking at your[ unix domain socket `Notify` implementation in the `kqueue`](https://github.com/smol-rs/polling/blob/master/src/kqueue.rs#L356), it is pretty much identical in behavior to the pipe-based notification mechanism in the `poll` module, except based on unix domain sockets. Shall I re-base the one in `poll` that would be `#[cfg(not(target_os = "espidf"))]` to be also based on unix domain sockets (not sure these are available everywhere else?) or shall I keep it based on pipes, as it is now?
* If we go for domain sockets, do you want the domain socket impl somehow centralized and outside of `kqueue` and `poll` (not sure that's easily possible but I can try), or do you want to separate implementations - one in `kqueue` and one in `poll`?

I'm in favor of a minimal variant where `poll` still has its own, private implementation based on `pipe` (and `eventfd` for the ESP IDF) as this stuff is difficult to test with so many variants of polling and operating systems, but also would like to understand where you want to drive this in the end, so that I don't have to redo the PR multiple times.

notgull:
- Unfortunately `UnixStream` does not work on Fuchsia. As Fuchsia isn't technically Unix, `libstd` does not expose it. So pipes are necessary for this use case.
- I would prefer to keep the `Notifier` type in this module. The goal is to have each module be as self contained as possible. Not to mention, the code can't really be shared between Unix and Windows.

ivmarkov:
> * Unfortunately `UnixStream` does not work on Fuchsia. As Fuchsia isn't technically Unix, `libstd` does not expose it. So pipes are necessary for this use case.
> * I would prefer to keep the `Notifier` type in this module. The goal is to have each module be as self contained as possible. Not to mention, the code can't really be shared between Unix and Windows.

OK it should be ready for another review.
* I have run the unit tests for both the `pipe` based approach as well as the `eventfd` based one on Linux as it supports both approaches.
* The one weird thing I found is that Linux really wants the eventfd handle to be registered with `PollFlags::IN` - as documented actually, while ESP IDF doesn't care.

notgull (changes requested):
Overall this looks good to me. Can you add a test to CI to make sure this code builds when targeting ESP-IDF? Even a simple `cargo check` would be enough.

Ignore the Wine and Cross failures in CI, they aren't your fault.

ivmarkov:
> Overall this looks good to me. Can you add a test to CI to make sure this code builds when targeting ESP-IDF? Even a simple `cargo check` would be enough.
> 
> Ignore the Wine and Cross failures in CI, they aren't your fault.

`cargo check` should be doable, yes, but with the following two caveats:
* Caveat 1: It will be a rust `nightly`-only exercise. The problem is, ESP IDF - being a Tier 3 target - does not get a pre-built STD by default. So we do need a command line like `rustup component add rust-src --toolchain nightly; cargo +nightly check --target riscv32imc-esp-espidf -Zbuild-std=std,panic_abort`, where the `-Zbuild-std` part is unfortunately a nightly-only feature of Cargo. 
* Caveat 2: The `eventfd` binding in `rustix` for ESP IDF specifically, just got merged [the other day](https://github.com/bytecodealliance/rustix/pull/769), and thus missed the 0.38.6 release, so it is only in `rustix`'s mainline. Let me actually open a PR against their repo for issuing another `rustix` point release

So if you are OK with 'Caveat 1`, yes, I'll work on that.

ivmarkov:
OK, CI is ready. For now (and until `rustix` 0.38.7 is released), `cargo check` for ESP IDF runs with the `rustix` dependency patched against its mainline (and there is a weird warning that the patched dependency is not used, but surely it is or else the build would've failed).

notgull (approved):
Thanks!

NetBSD failure looks spurious to me, not sure why that's happening.

PR #129, opened 2023-08-04
Partial support for Rustix 0.38.x

daggerrz:
Rustix 0.38.x has made breaking changes to the event APIs by splitting some of the modules. See https://github.com/bytecodealliance/rustix/commit/29b07bbdac1c7d704c836db337c13928b47bcdf3

This PR updates the implementations for `kqueue` and `epoll` to use the new module structure. I tried to update `poll` as well, but the various `pipe`, `pipe_with` and similar functions now seem to be private (in the backend only) and that's beyond my basic understanding of how this works, unfortunately.


notgull:
Porting to Rustix v0.38 was already done in #123.

Thanks anyways!

PR #131, opened 2023-08-06
Remove patch.crates-io now that rustix 0.38.7 is out

ivmarkov:
Subject says it all - rustix.0.38.7 [just got released](https://github.com/bytecodealliance/rustix/pull/770).


taiki-e (approved):
Thanks.

PR #132, opened 2023-08-06
Allow pinning Rust nightly to concrete version

ivmarkov:
A nit really: replace `==` with `startWith` so that if one day we decide to pin `nightly` to a specific version (i.e. `nightly-2023-08-01` the Tier 3 CI would still run.


taiki-e (approved):
Thanks.

PR #133, opened 2023-08-09
Extract the Events struct and make the Event struct opaque

### LINKED ISSUE #66: Expose the Events structure instead of using Vec<Event>

### LINKED ISSUE #112: Enable HUP tracking in Event
It would be nice to track events aside from readable/writable in `Event`. This would involve having a private storage field in the `Event` type indicating other readiness types, like HUP, Inode flags, or process flags.

Requires a breaking change.

notgull:
- Extracts the `Events` struct from `Poller`. This allows us to avoid an intermediate allocation when polling for events in `async-io`.
- Makes `Event` opaque, containing platform-specific data. Closes #66.
- Adds `HUP` and `PRI` tracking to `Event` that is platform-dependent. Closes #112 and [this issue](https://github.com/alacritty/alacritty/pull/6846/files#r1277724556).

### DIFF
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -59,11 +59,7 @@ impl Poller {
 
             poller.add(
                 poller.event_fd.as_raw_fd(),
-                Event {
-                    key: crate::NOTIFY_KEY,
-                    readable: true,
-                    writable: false,
-                },
+                Event::readable(crate::NOTIFY_KEY),
                 PollMode::Oneshot,
             )?;
         }
@@ -177,11 +173,7 @@ impl Poller {
             // Set interest in timerfd.
             self.modify(
                 timer_fd.as_fd(),
-                Event {
-                    key: crate::NOTIFY_KEY,
-                    readable: true,
-                    writable: false,
-                },
+                Event::readable(crate::NOTIFY_KEY),
                 PollMode::Oneshot,
             )?;
         }
@@ -213,11 +205,7 @@ impl Poller {
         let _ = read(&self.event_fd, &mut buf);
         self.modify(
             self.event_fd.as_fd(),
-            Event {
-                key: crate::NOTIFY_KEY,
-                readable: true,
-                writable: false,
-            },
+            Event::readable(crate::NOTIFY_KEY),
             PollMode::Oneshot,
         )?;
         Ok(())
@@ -322,6 +310,7 @@ impl Events {
                 key: ev.data.u64() as usize,
                 readable: flags.intersects(read_flags()),
                 writable: flags.intersects(write_flags()),
+                extra: EventExtra { flags },
             }
         })
     }
@@ -334,5 +323,20 @@ impl Events {
     /// Get the capacity of the list.
     pub fn capacity(&self) -> usize {
         self.list.capacity()
+    }
+}
+
+/// Extra information about this event.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EventExtra {
+    flags: epoll::EventFlags,
+}
+
+impl EventExtra {
+    /// Create an empty version of the data.
+    pub fn empty() -> EventExtra {
+        EventExtra {
+            flags: epoll::EventFlags::empty(),
+        }
     }
 }
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -308,9 +308,9 @@ unsafe impl Send for Events {}
 
 impl Events {
     /// Creates an empty list.
-    pub fn new() -> Events {
+    pub fn with_capacity(cap: usize) -> Events {
         Events {
-            list: epoll::EventVec::with_capacity(1024),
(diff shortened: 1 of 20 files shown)


src/iocp/mod.rs
```diff
@@ -622,16 +622,21 @@ unsafe impl Send for Events {}
...
 
     /// Iterate over I/O events.
     pub(super) fn iter(&self) -> impl Iterator<Item = Event> + '_ {
         self.packets.iter().copied()
     }
+
+    /// Clear the list of events.
+    pub(super) fn clear(&mut self) {
+        self.packets.clear();
+    }
```
fogti, line 639: why are these `pub(super)`?

src/epoll.rs
```diff
@@ -339,4 +339,24 @@ impl EventExtra {
...
+
+    /// Tell if the interrupt flag is set.
+    pub fn is_hup(&self) -> bool {
+        self.flags.contains(epoll::EventFlags::HUP)
+    }
+
+    /// Tell if the priority flag is set.
+    pub fn is_pri(&self) -> bool {
+        self.flags.contains(epoll::EventFlags::PRI)
+    }
```
fogti, line 361: It might be a good idea to mark these as `#[inline]`...

fogti: also, why doesn't `set_*` take a boolean whether to enable/disable the flag?

tests/other_modes.rs
```diff
@@ -229,10 +229,9 @@ fn edge_oneshot_triggered() {
         .wait(&mut events, Some(Duration::from_secs(10)))
         .unwrap();
 
-    assert_eq!(
-        events.iter().collect::<Vec<_>>(),
-        [Event::readable(reader_token)]
-    );
+    assert_eq!(events.len(), 1);
+    assert!(events.iter().next().unwrap().readable);
+    assert!(!events.iter().next().unwrap().writable);
```
fogti, line 236: this pattern occurs multiple times in this commit but it looks like a code smell. imo it would be better find a way to do both of these asserts on the same flag variable. (e.g. by retrieving `events.iter().next().unwrap()` before these asserts and reuse the result)

fogti:
Imo it would be a good idea to squash all the doctest fixups before merging.

src/iocp/mod.rs
notgull: Fixed to be `pub`

src/iocp/mod.rs
```diff
@@ -698,9 +783,14 @@ impl PacketUnwrapped {
                 socket.mode = mode;
                 socket.interest_error = true;
 
+                // If there was a change, indicate that we need an update.
                 match socket.status {
-                    SocketStatus::Polling { readable, writable } => {
-                        (interest.readable && !readable) || (interest.writable && !writable)
+                    SocketStatus::Polling { flags } => {
+                        let our_flags =
+                            event_to_afd_mask(interest.readable, interest.writable, true)
```
fogti, line 790: why is this given `true` as the last argument?

src/iocp/mod.rs
```diff
@@ -780,12 +877,16 @@ impl PacketUnwrapped {
...
         // Check the current status.
         match socket.status {
-            SocketStatus::Polling { readable, writable } => {
+            SocketStatus::Polling { flags } => {
                 // If we need to poll for events aside from what we are currently polling, we need
                 // to update the packet. Cancel the ongoing poll.
-                if (socket.interest.readable && !readable)
-                    || (socket.interest.writable && !writable)
-                {
+                let our_flags = event_to_afd_mask(
```
fogti, line 883: it is probably a good idea to introduce a small wrapper around `event_to_afd_mask` that just takes such a `socket.interest` and does produce such a `our_flags`, as this pattern repeats several times...

src/iocp/mod.rs
notgull: Because `interest_error` in this case is always `true`. I've clarified it here.

fogti (approved):
LGTM

PR #134, opened 2023-08-10
Add a pipe-based notifier to epoll

### LINKED ISSUE #122: Add a fallback to Linux to avoid eventfd if it fails on Gramine
cc https://github.com/tokio-rs/mio/issues/1682

For reference: https://gramine.readthedocs.io/en/stable/manifest-syntax.html#allowing-eventfd

notgull:
In some containers, eventfd is not available as it cannot be implemented securely in some hosts. This commit adds a fallback notifier that uses a pipe instead of eventfd.

Closes #122

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,11 @@ jobs:
           # Note: This cfg is intended to make it easy for polling developers to test
           # the backend that uses poll, and is not a public API.
           RUSTFLAGS: ${{ env.RUSTFLAGS }} --cfg polling_test_poll_backend
+      - run: cargo test
+        env:
+          # Note: This cfg is intended to make it easy for polling developers to test
+          # the backend that uses pipes, and is not a public API.
+          RUSTFLAGS: ${{ env.RUSTFLAGS }} --cfg polling_test_epoll_pipe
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -55,6 +55,12 @@ jobs:
           # the backend that uses poll, and is not a public API.
           RUSTFLAGS: ${{ env.RUSTFLAGS }} --cfg polling_test_poll_backend
         if: startsWith(matrix.os, 'ubuntu')
+      - run: cargo test
+        env:
+          # Note: This cfg is intended to make it easy for polling developers to test
+          # the backend that uses pipes, and is not a public API.
+          RUSTFLAGS: ${{ env.RUSTFLAGS }} --cfg polling_test_epoll_pipe
```
taiki-e, line 62: Is this really a cfg mainly intended for use in testing?
If not, it may be preferable to remove "\_test\_" and simply state that it is an unstable API.

notgull: Yes, it should only be used in testing. The pipe fallback is mostly for containers and should happen dynamically.

taiki-e (approved):
Thanks!

PR #136, opened 2023-08-16
Specify behavior when registered in multiple pollers

### LINKED ISSUE #81: Specify behavior when sources are added to multiple Pollers or Poller-like constructs
As far as I can tell, from some testing that I just ran, when a source is registered into multiple `Poller`s, any events that originate from the source will appear on every `Poller`. Do we want to specify this behavior as guaranteed, or do we want to go in another direction.

The reason why I ask is because, as of Windows 8.1, there is a [`WSAEventSelect`](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaeventselect) function that is stable, documented and would require no external dependencies over the current `wepoll` backend. However, it comes with the caveat that every socket can only be registered in one `WSAEVENT` at a time. If we decide that we want to guarantee the current general behavior, we could not switch to this.

Interaction with `mio` isn't an issue, since `mio` says that being registered to `mio::Poll` [is exclusive](https://docs.rs/mio/latest/mio/struct.Poll.html#accessing-raw-fdsockethandle).

notgull:
This adds documentation to add() describing what happens when a source is registered in multiple pollers. A test is also added to ensure this behavior.

Closes #81

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -411,6 +411,17 @@ impl Poller {
     /// [`modify()`][`Poller::modify()`] again after an event is delivered if we're interested in
     /// the next event of the same kind.
     ///
+    /// It is possible to register interest in the same file descriptor or socket using multiple
+    /// separate [`Poller`] instances. When one event is delivered, all [`Poller`] instances will
+    /// be notified. The only exception is `Oneshot` mode. Due to underlying platform limitations,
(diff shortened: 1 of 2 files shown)


notgull:
Huh, weird, apparently that doesn't work for Windows? Not sure if that is an AFD implementation detail or not. Oh well, it's easy enough to paper over.

PR #119, opened 2023-06-11
Replace log with tracing

notgull:
Supersedes #65

PR #139, opened 2023-08-20
v3.0.0

notgull:
- Replace `libc` in all backends with the `rustix` crate (#108, #146).
- Use `tracing` instead of `log` for logging (#119).
- **Breaking:** Rework the API to use I/O safety. Note that this makes several previously safe functions unsafe. (#123)
- Add support for the ESP-IDF platform. (#128)
- **Breaking:** Make `Event` partially opaque, and create a new `Events` struct for holding events. (#133)
- Add support for running `polling` in Linux containers without `eventfd` available. (#134)
- Specify the behavior when registered in multiple `Poller`s. (#136)
- **Breaking:** Use `c_int` from the standard library in `polling::os::kqueue` instead of defining our own. (#143)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# Version 3.0.0
+
+- Replace `libc` in all backends with the `rustix` crate (#108).
+- Use `tracing` instead of `log` for logging (#119).
+- **Breaking:** Rework the API to use I/O safety. Note that this makes several previously safe functions unsafe. (#123)
+- Add support for the ESP-IDF platform. (#128)
+- **Breaking:** Make `Event` partially opaque, and create a new `Events` struct for holding events. (#133)
+- Add support for running `polling` in Linux containers without `eventfd` available. (#134)
+- Specify the behavior when registered in multiple `Poller`s. (#136)
+- **Breaking:** Use `c_int` from the standard library in `polling::os::kqueue` instead of defining our own. (#143)
+- **Breaking:** Remove the useless `std` feature. (#147)
+
 # Version 2.8.0
 
 - Add functionality for posting events to the IOCP. (#101)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,9 +2,9 @@
 name = "polling"
 # When publishing a new version:
 # - Update CHANGELOG.md
-# - Create "v2.x.y" git tag
-version = "2.8.0"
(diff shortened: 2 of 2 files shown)


notgull:
@smol-rs/admins Any objections to me publishing this? I do not foresee needing to make any more breaking changes after this.

zeenix:
None from me.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# Version 3.0.0
+
+- Replace `libc` in all backends with the `rustix` crate (#108).
+- Use `tracing` instead of `log` for telemetry (#119).
```
taiki-e, line 4: Personally, I prefer "logging" to "telemetry" here. While "telemetry" can mean "from remote", that's not what we're doing.

taiki-e (approved):
LGTM aside from nits on changelog.

taiki-e (approved):

fogti:
why is `libc` still in `dev-dependencies`?

fogti:
why is `rustix` pulled in with `"std"` as a feature?

notgull:
> why is `libc` still in `dev-dependencies`?

Good question, I've opened #146 to address it.


> why is `rustix` pulled in with `"std"` as a feature?

This is necessary for `std::os::fd::BorrowedFd` to be used in `rustix`.

fogti:
hmm why does this crate even have an `std` feature?

notgull:
> hmm why does this crate even have an `std` feature?

Good question. This was added in 272bb11eafd22c063bc2e7825070c21443a8fc78 for reasons that elude me; maybe it was meant to port this crate to not require `libstd` at some point. Although `std::os::fd::*` is necessary to register sources, so it doesn't really seem all too relevant to me.

I can remove it.

fogti:
Remove it.

notgull:
> Remove it.

Ok, done in https://github.com/smol-rs/polling/pull/147

fogti:
ok, ChangeLog should be updated to include #147.
was the "std" feature present in some proper release?

notgull:
> ok, ChangeLog should be updated to include #146 and #147. was the "std" feature present in some proper release? Yes, since v2.3.0.

#146 only affected dev dependencies, so it shouldn't be put in the user facing changelog. I added #147 though.

fogti (approved):

PR #140, opened 2023-08-21
feat: enable construct event constantly

tisonkun:
Maybe https://github.com/smol-rs/polling/pull/133 can prevent this from being possible; since it includes sys details for `Event`.

These `sys` things are one of the main complain on `mio` that I want to get rid of.

Compare with `mio`, it can obtain a `const TOKEN: Token = Token(0)`, while in `polling` we cannot obtain a const interest `Event` like `const INTEREST: Event = Event::readable(1)`. I don't want to know about the `sys` things.

cc @notgull

### DIFF
--- a/src/iocp/mod.rs
+++ b/src/iocp/mod.rs
@@ -651,7 +651,7 @@ pub struct EventExtra {
 impl EventExtra {
     /// Create a new, empty version of this struct.
     #[inline]
-    pub fn empty() -> EventExtra {
+    pub const fn empty() -> EventExtra {
         EventExtra {
             flags: AfdPollMask::empty(),
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -186,7 +186,7 @@ impl Event {
     /// All kinds of events (readable and writable).
     ///
     /// Equivalent to: `Event { key, readable: true, writable: true }`
-    pub fn all(key: usize) -> Event {
+    pub const fn all(key: usize) -> Event {
         Event {
             key,
             readable: true,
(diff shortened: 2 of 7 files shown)


tisonkun:
Perhaps I can use a `const KEY: usize = 1` and `Event::readable(KEY)` and that should be fine.

Keep this PR open for ideas on whether we can make the event constructors themselves const.

src/iocp/mod.rs
```diff
@@ -651,7 +651,7 @@ pub struct EventExtra {
 impl EventExtra {
     /// Create a new, empty version of this struct.
     #[inline]
-    pub fn empty() -> EventExtra {
+    pub const fn empty() -> EventExtra {
         EventExtra {
             flags: AfdPollMask::empty(),
         }
```
notgull, line 657: Fix the compile error by also making `AfdPollMask::empty()` constant.

src/lib.rs
```diff
@@ -186,7 +186,7 @@ impl Event {
     /// All kinds of events (readable and writable).
     ///
     /// Equivalent to: `Event { key, readable: true, writable: true }`
-    pub fn all(key: usize) -> Event {
+    pub const fn all(key: usize) -> Event {
```
notgull, line 189: You forgot to make the `poll.rs` version of `EventEmpty` `const`.

notgull (changes requested):
I think this would be a good addition to this crate.

tisonkun:
Updated. Let's see if all the empty branches can be `const`.

tisonkun:
All tests passed. PTAL @notgull

notgull (approved):
Looks good to me!

PR #143, opened 2023-08-29
Use std::os::raw::c_int and remove our own type alias

taiki-e:
They are always i32 in macos/bsd, but it is better to be clear that they are the same as std c_int.


taiki-e:
Or, we can just remove this alias from this module by using std c_int, as part of the v3 breaking changes.

notgull (approved):
Thanks!

Issue #87, opened 2023-02-14
Support no_std mode


notgull:
The backend that `polling` uses right now, `libc`, and the backend we intend to use in the future, `rustix`, both support `no_std` mode. It would be nice for #86 to be able to support `no_std`, if only so that the entire Rust standard library isn't bunded into the library.

This would be a breaking change, since we currently use `io::Error` in the public API.

taiki-e:
Given that our code uses std-specific items such as Mutex, I'm not sure if this is actually possible or reasonable.

notgull:
Ah, yeah, I just noticed that. Closing this issue.

PR #147, opened 2023-09-04
Remove the std default feature

notgull:
Added in 272bb11eafd22c063bc2e7825070c21443a8fc78 for reasons that are unclear to me. It serves no purpose aside from disabling the API of the entire crate, so it's best to remove it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,6 @@ exclude = ["/.*"]
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 
-[features]
-default = ["std"]
-std = []
-
 [dependencies]
 cfg-if = "1"
 tracing = { version = "0.1.37", default-features = false }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,8 +53,6 @@
 //! # std::io::Result::Ok(())
 //! ```
 
-#![cfg(feature = "std")]
-#![cfg_attr(not(feature = "std"), no_std)]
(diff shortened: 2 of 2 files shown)


fogti (approved):

fogti:
what about the CI change in the original commit?

notgull:
> what about the CI change in the original commit? (I checked now, it's no longer present)

Looks like it was removed in https://github.com/smol-rs/polling/commit/dec94cb423489e2eec27295fcb89f285a380536b when taiki-e was cleaning up the CI configuration

taiki-e:
This was necessary to add no-std support without requiring a breaking change. (https://github.com/crossbeam-rs/crossbeam/pull/508#issuecomment-631458481 has a more detailed explanation of this.)

However, given that we have decided not to do https://github.com/smol-rs/polling/issues/87, that may no longer be necessary.


> > what about the CI change in the original commit? (I checked now, it's no longer present)
> 
> Looks like it was removed in [dec94cb](https://github.com/smol-rs/polling/commit/dec94cb423489e2eec27295fcb89f285a380536b) when taiki-e was cleaning up the CI configuration

That check was not just removed, but replaced with a more robust [`cargo hack build --feature-powerset --no-dev-deps`](https://github.com/smol-rs/polling/commit/dec94cb423489e2eec27295fcb89f285a380536b#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR32).

PR #149, opened 2023-09-11
introduce Event::new()

### LINKED ISSUE #148: awkward to construct `Event` in 3.0.0
we have some constructors eg
```rust
 impl Event {
      /// All kinds of events (readable and writable).
      ///
      /// Equivalent to: `Event { key, readable: true, writable: true }`
      pub const fn all(key: usize) -> Event {
          ...
```
but the claimed equivalence is not true: the addition of the private `extra` field makes it impossible to construct an `Event` that way.

So code that in polling 2.x goes like:
```
  let readable = am_i_interested_in_reads();
  let writable = am_i_interested_in_writes();
  let event = Event { key, readable, writable };
```
has become awkward to write in polling 3.0.0

can we get an `Event::new(key, readable, writable)` to recover this?  or failing that I guess treat this as a bug report against the no-longer-true docs.

dimbleby:
fixes #148

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,52 +181,46 @@ pub enum PollMode {
 }
 
 impl Event {
-    /// All kinds of events (readable and writable).
-    ///
-    /// Equivalent to: `Event { key, readable: true, writable: true }`
-    pub const fn all(key: usize) -> Event {
+    /// Create a new event.
+    pub const fn new(key: usize, readable: bool, writable: bool) -> Event {
         Event {
             key,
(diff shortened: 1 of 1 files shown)


notgull (changes requested):
Could you make all of the functions that just wrap around `new()` `#[inline]`?

dimbleby:
> Could you make all of the functions that just wrap around `new()` `#[inline]`?

done

notgull (approved):
Thanks!

PR #157, opened 2023-10-04
chore: prefer usize::MAX to std::usize::MAX

tisonkun:
Per https://github.com/smol-rs/polling/pull/140#issue-1859612645

Hopefully, we can move `Event`s free. Open for comments :D

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,6 @@ use std::num::NonZeroUsize;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 use std::time::Duration;
-use std::usize;
 
 use cfg_if::cfg_if;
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -136,6 +135,9 @@ pub struct Event {
     extra: sys::EventExtra,
 }
 
+unsafe impl Send for Event {}
+unsafe impl Sync for Event {}
```
notgull, line 139: I don't like the additional `unsafe` here. Can we just add a test to make sure the `Event` struct is `Send` and `Sync`?

tisonkun: > add a test

Do you mean that we ensure the auto trait happens by testing? What's the difference to this patch that let the compiler do the "test"?

These two lines are automatically done if it meets the auto trait conditions. We declare them explicitly for testing here.

notgull: Like this: https://github.com/smol-rs/async-executor/blob/2cfb6e4ed03c7eaf6c5fb2d3170ac870ba0a9b9d/src/lib.rs#L1065

tisonkun: OK. I'm pushing a commit now - this seems a style preference :D

tisonkun: We already have it in `_assert_send_and_sync`.

Then I'll convert this patch as a trivial code tidy one.

notgull (approved):
Thanks!

PR #159, opened 2023-10-08
Remove needless imports

taiki-e:
I missed them in https://github.com/smol-rs/polling/pull/158.


zeenix:
> missed them in #158

Hence why you might want to give people a chance to review PRs before merging. 😊

Thanks for the work though! 👍🙏

PR #163, opened 2023-10-18
Fix wrong link in docs of Poller::wait()

psychon:
Once upon a time, this got a Vec as an argument, but that was replaced with the Events struct. Thus, this should link to Events and not Vec.

Screenshot of result:  
foo


notgull (approved):
Thanks!

PR #164, opened 2023-10-21
Handle interrupts while polling

### LINKED ISSUE #162: Should this crate intercept Interrupt errors?
The polling system call can be interrupted in some cases. However, this error is bubbled up to the user. `async-io` handles it explicitly, but other users may not expect to need to handle this error. Should we be handling it on the `polling` side instead?

cc https://github.com/psychon/x11rb/issues/892

notgull:
Previous, `Poller::wait` would bubble signal interruption error to the user. However, this may be unexpected for simple use cases. Thus, this commit makes it so, if `ErrorKind::Interrupted` is received by the underlying `wait()` call, it clears the events and tries to wait again.

This also adds a test for this interruption written by @psychon.

Closes #162

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -646,7 +646,7 @@ impl Poller {
     /// poller.delete(&socket)?;
     /// # std::io::Result::Ok(())
     /// ```
-    pub fn wait(&self, events: &mut Events, mut timeout: Option<Duration>) -> io::Result<usize> {
+    pub fn wait(&self, events: &mut Events, timeout: Option<Duration>) -> io::Result<usize> {
         let span = tracing::trace_span!("Poller::wait", ?timeout);
         let _enter = span.enter();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -656,6 +656,7 @@ impl Poller {
                 if let Err(e) = self.poller.wait(&mut events.events, timeout) {
                     // If the wait was interrupted by a signal, clear events and try again.
                     if e.kind() == io::ErrorKind::Interrupted {
+                        events.clear();
                         continue;
                     } else {
                         return Err(e);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -651,14 +651,23 @@ impl Poller {
         let _enter = span.enter();
 
         if let Ok(_lock) = self.lock.try_lock() {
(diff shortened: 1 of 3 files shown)


psychon:
I would claim that this mis-handles timeouts. A poll with a timeout of one second can now run infinitely long (although this is quite unlikely):

- `self.poller.wait(events, OneSecond)` runs for 0.5 seconds
- A signal arrives and interrupts the poller
- `self.poller.wait(events, OneSecond)` runs for 0.5 seconds
- A signal arrives and interrupts the poller
- repeat indefinitely

I would propose:
```diff
diff --git a/src/lib.rs b/src/lib.rs
index 9882c8e..8078ee1 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -652,7 +652,14 @@ impl Poller {
 
         if let Ok(_lock) = self.lock.try_lock() {
             // Wait for I/O events.
-            self.poller.wait(&mut events.events, timeout)?;
+            match self.poller.wait(&mut events.events, timeout) {
+                Ok(()) => {}
+                // Turn interruption into a spurious wakeup without errors
+                Err(e) if e.kind() == io::ErrorKind::Interrupted => {
+                    events.clear();
+                },
+                Err(e) => return Err(e),
+            };
 
             // Clear the notification, if any.
             self.notified.swap(false, Ordering::SeqCst);
```

src/lib.rs
```diff
@@ -651,14 +651,24 @@ impl Poller {
...
-
-            // Indicate number of events.
-            Ok(events.len())
+            loop {
+                // Wait for I/O events.
+                if let Err(e) = self.poller.wait(&mut events.events, timeout) {
+                    // If the wait was interrupted by a signal, clear events and try again.
+                    if e.kind() == io::ErrorKind::Interrupted {
+                        events.clear();
+                        continue;
```
fogti, line 660: this branch should adjust the timeout.

fogti:
An alternative would be to only do this if no timeout was given (because spurious/non-action wakeups need to be handled anyways when using timeouts usually)

fogti (changes requested):
(timeout concerns)

fogti (approved):

PR #171, opened 2023-12-26
Add add_outlives method to Poller

### LINKED ISSUE #161: Add a way to add a source that outlives the Poller
If a specific file descriptor or source outlives the `Poller`, it can be `add`ed to it without any concerns of whether or not it will be dropped. It would be nice to have an `add_outlives()` function that takes a source that is guaranteed to outlive the `Poller`, through lifetimes. While this doesn't work for the common use cases (`async-io` and `calloop`), for simple, shorter scale use cases this could be used to avoid unsafe code (`alacritty`).

cc https://github.com/psychon/x11rb/issues/891

BrankoKrstic:
resolves #161 

The previous implementation only implemented unsafe methods for registering file descriptors with `Poller`. This implements an `add_outlives` method that can guarantee safety by accepting a reference to an `AsSource` which outlives the `Poller`. Since file descriptors are guaranteed to be valid through the lifetime of the `Poller`, there is no need to ever delete to maintain soundness.

Please advise if I should change anything.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -349,13 +349,15 @@ impl Event {
 }
 
 /// Waits for I/O events.
-pub struct Poller {
+pub struct Poller<'a> {
     poller: sys::Poller,
     lock: Mutex<()>,
     notified: AtomicBool,
+    // Being contravariant over 'a requires borrowed file descriptors outlive the `Poller`
+    _marker: PhantomData<fn(&'a ())>,
 }
 
-impl Poller {
+impl<'a> Poller<'a> {
     /// Creates a new poller.
     ///
     /// # Examples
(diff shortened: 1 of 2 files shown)


notgull (changes requested):
This is a breaking change, as `Poller` now has a lifetime parameter. I was envisioning something like this:

```rust
pub fn add_outlives<'poll, 'src, T: AsSource>(&'poll self, source: &'src T, interest: Event) -> io::Result<()>
where 'src: 'poll
 { /* ... */ }
```

This way the reference to the source is guaranteed to outlive the `Poller`.

BrankoKrstic:
I tried this initially, but I don't think it will work if the source lifetime is not somehow tied to the Poller struct. The borrow checker will just shorten the poller lifetime inside `add_outlives`.

With the above example this compiles:
```rust
let poller = Poller::new().unwrap();

{
    let file = File::open("./file.txt").unwrap();
    let _ = poller.add_outlives(&file, Event::all(7));
    drop(file);
}
println!("{:?}", poller);
```

notgull:
That's disappointing. I was under the impression that the borrow checker would catch this.

Still, thanks for checking!

PR #180, opened 2024-01-16
Add a way to wait on process by PID

notgull:
This is needed to make processes work in `async-process`.

### DIFF
--- a/src/os/kqueue.rs
+++ b/src/os/kqueue.rs
@@ -173,7 +173,7 @@ impl Filter for Signal {}
 #[derive(Debug)]
 pub struct Process<'a> {
     /// The process ID to monitor.
-    pid: std::os::raw::c_int,
+    pid: rustix::process::Pid,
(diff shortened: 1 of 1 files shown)


src/os/kqueue.rs
```diff
@@ -171,11 +172,14 @@ impl Filter for Signal {}
 /// Monitor a child process.
 #[derive(Debug)]
 pub struct Process<'a> {
-    /// The child process to monitor.
-    child: &'a Child,
+    /// The process ID to monitor.
+    pid: std::os::raw::c_int,
```
fogti, line 176: why don't we keep the type information that `pid` is non-zero? Why not just save a `rustix::process::Pid`, it isn't part of the exported API so I suppose it wouldn't hurt and get rid of most of the weird handling here...

notgull: Good point, fixed

Issue #184, opened 2024-01-25
Does not have access to EventExtra


irvingoujAtDevolution:
When doing TCP connection with IOCP, if the other end have no TCP server listening, the extra event arg will have a flag CONNECT_FAIL, but current API does not provide access to this arg.

notgull:
The inner field here is intentionally obscured. It should only be used from the `Event` structure's public APU.

PR #185, opened 2024-01-25
feat : add the ability to identify if tcp connection has failed

irvingoujAtDevolution:
Added is_connection_failed method to identify if a tcp connection has failed, please see example `tcp_client.rs`. 
Let me know if adding socket2 to dev-dependency is acceptable, I will remove it if not.

address https://github.com/smol-rs/polling/issues/184

### DIFF
--- a/examples/tcp_client.rs
+++ b/examples/tcp_client.rs
@@ -11,30 +11,23 @@ fn main() -> io::Result<()> {
     }
     let addr = net::SocketAddr::new(net::Ipv4Addr::LOCALHOST.into(), 8080);
     socket.set_nonblocking(true)?;
-    let res = socket.connect(&addr.into());
+    let _ = socket.connect(&addr.into());
 
     let mut events = polling::Events::new();
-    // while let Err(ref e) = res {
-        // if e.kind() != io::ErrorKind::WouldBlock {
-        //     return Err(io::Error::new(e.kind(), e.to_string()));
-        // }
 
-        events.clear();
-        poller.wait(&mut events, None)?;
+    events.clear();
+    poller.wait(&mut events, None)?;
 
-        let event = events.iter().next();
-        if event.is_none() {
-            println!("no event");
-            // break;
-        }
+    let event = events.iter().next();
+    let Some(event) = event else {
+        println!("no event");
+        return Ok(());
+    };
 
-        let event = event.unwrap();
-        println!("event: {:?}", event);
-        if event.is_connect_failed() {
-            println!("connect failed");
-            // break;
-        }
-    // }
+    println!("event: {:?}", event);
+    if event.is_connect_failed() {
+        println!("connect failed");
+    }
 
     Ok(())
 }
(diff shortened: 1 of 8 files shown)


src/kqueue.rs
```diff
@@ -371,6 +371,11 @@ impl EventExtra {
     pub fn is_pri(&self) -> bool {
         false
     }
+
+    #[inline]
+    pub fn is_connect_failed(&self) -> bool {
+        unimplemented!("is connect failed is not supported on kqueue");
```
notgull, line 377: Just return `false` here, no need to panic

examples/tcp_client.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut events = polling::Events::new();
+
+    events.clear();
+    poller.wait(&mut events, None)?;
+
+    let event = events.iter().next();
+    let Some(event) = event else {
+        println!("no event");
+        return Ok(());
+    };
```
notgull, line 25: Is this syntax compatible with Rust 1.63?

src/lib.rs
```diff
@@ -332,6 +332,17 @@ impl Event {
         self.extra.is_pri()
     }
 
+    /// Tell if this event is the result of a connect failure.
+    ///
+    /// This indicates a tcp connection has failed, it corresponds to the `EPOLLERR` along with `EPOLLHUP` event in linux
+    /// and `CONNECT_FAILED` event in windows IOCP.
+    ///
```
notgull, line 339: This should have an example.

src/poll.rs
```diff
@@ -426,6 +426,13 @@ impl EventExtra {
     pub fn is_pri(&self) -> bool {
         self.flags.contains(PollFlags::PRI)
     }
+
+    #[inline]
+    pub fn is_connect_failed(&self) -> bool {
+        // need reviewer's special attention, as I do not have access to a system that supports this
+        // this is a guess based on the documentation of `poll()`
+        self.flags.contains(PollFlags::ERR) || self.flags.contains(PollFlags::HUP)
```
notgull, line 434: This looks right to me

notgull (changes requested):
There should be a function on `Poller` indicating whether or not `is_connect_failed` is supported, like with edge triggered polling.

src/kqueue.rs
irvingoujAtDevolution: updated

examples/tcp_client.rs
irvingoujAtDevolution: updated

src/lib.rs
irvingoujAtDevolution: added

notgull:
Actually, I think it might be a saner model to have it return `Option<bool>` and return `None` for BSD.

irvingoujAtDevolution:
> Actually, I think it might be a saner model to have it return `Option<bool>` and return `None` for BSD.

You mean for is_connect_failed? 
if we don't support it, we return None, if we support it,we return the value?

do we still need supports_is_connect_failed in that case?

irvingoujAtDevolution:
> Yes, that's right. No we do not.

updated!

notgull (approved):
Thanks!

PR #187, opened 2024-01-28
v3.3.3

notgull:
- On ESP-IDF, there is an EPERM error when initializing the `Poller` without first having initialized the ESP-IDF eventfd subsystem. Now, when this happens the error will mention the need to call "esp_vfs_eventfd_register" first. (#186)


notgull:
Closed in favor of #188

PR #189, opened 2024-02-07
Fix: replace is_connect_failed with is_err

irvingoujAtDevolution:
In linux, epoll, EPOLLHUP may happen even if no connection call is made. It would confuse callers for what is actually happening.
Replaced is_connect_failed, and we detect if connection failed by using the combination of is_err and is_interrupt, please see the example, tcp_client

### DIFF
--- a/examples/tcp_client.rs
+++ b/examples/tcp_client.rs
@@ -1,0 +1,36 @@
+use std::{io, net};
+
+use polling::Event;
+use socket2::Type;
+
+fn main() -> io::Result<()> {
+    let socket = socket2::Socket::new(socket2::Domain::IPV4, Type::STREAM, None)?;
+    let poller = polling::Poller::new()?;
+    unsafe {
+        poller.add(&socket, Event::new(0, true, true))?;
+    }
+    let addr = net::SocketAddr::new(net::Ipv4Addr::LOCALHOST.into(), 8080);
+    socket.set_nonblocking(true)?;
+    let _ = socket.connect(&addr.into());
+
+    let mut events = polling::Events::new();
+
+    events.clear();
+    poller.wait(&mut events, None)?;
+
+    let event = events.iter().next();
+    let event = match event {
+        Some(event) => event,
+        None => {
+            println!("no event");
+            return Ok(());
+        }
+    };
+
+    println!("event: {:?}", event);
+    if event.is_connect_failed().unwrap_or_default() {
+        println!("connect failed");
+    }
+
+    Ok(())
+}
--- a/examples/tcp_client.rs
+++ b/examples/tcp_client.rs
@@ -1,3 +1,4 @@
+#![cfg(any(target_os = "linux", target_os = "windows"))]
 use std::io;
 
 #[cfg(target_os = "linux")]
--- a/examples/tcp_client.rs
+++ b/examples/tcp_client.rs
@@ -1,8 +1,5 @@
-#![cfg(any(target_os = "linux", target_os = "windows"))]
-use std::io;
-
 #[cfg(target_os = "linux")]
-fn main() -> io::Result<()> {
+fn main() -> std::io::Result<()> {
     use std::net;
     use std::{io::Write, time::Duration};
 
@@ -86,7 +83,7 @@ fn main() -> io::Result<()> {
 }
 
 #[cfg(target_os = "windows")]
-fn main() -> io::Result<()> {
+fn main() -> std::io::Result<()> {
     use polling::Event;
     use std::{io::Write, time::Duration};
 
@@ -151,7 +148,7 @@ fn main() -> io::Result<()> {
     Ok(())
 }
 
-#[cfg(any(target_os = "macos", target_os = "android"))]
+#[cfg(not(any(target_os = "linux", target_os = "windows")))]
 fn main() {
     println!("This example is not yet supported on this platform.");
 }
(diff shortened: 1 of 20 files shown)


notgull (changes requested):
Unfortunately we can't change `is_connect_failed` as that would be a breaking change. So we have to keep the name the same there.

irvingoujAtDevolution:
> Unfortunately we can't change `is_connect_failed` as that would be a breaking change. So we have to keep the name the same there.

I apologize for my previous mistake. I would like to have your opinion on whether it is better to add `is_err`, mark `is_connect_fail` as deprecated, or simply keep using the same name but change the implementation.

I would argue that the first approach is better, as it aligns with the semantics it provides.

notgull:
> I apologize for my previous mistake. I would like to have your opinion on whether it is better to add `is_err`, mark `is_connect_fail` as deprecated, or simply keep using the same name but change the implementation.
> 
> I would argue that the first approach is better, as it aligns with the semantics it provides.

I'd argue that the first would be better.

examples/tcp_client.rs
```diff
@@ -1,36 +1,155 @@
-use std::{io, net};
```
notgull, line 1: I would appreciate not changing this example and instead just wrapping the socket error bits in the configuration guards.

examples/tcp_client.rs
```diff
@@ -1,36 +1,155 @@
...
+
+    assert!(event.writable);
+    assert!(!event.is_interrupt());
+    assert!(!event.is_err().unwrap());
+
+    println!("event: {:?}", event);
+    println!("socket is now writable");
+    // ========================================================================
+    // the below is example of a bad socket
+    println!("testing bad socket");
```
notgull, line 55: If this is a functionality test it should be moved to be under the `tests` folder.

src/lib.rs
```diff
@@ -391,10 +391,63 @@ impl Event {
     /// Returns `Some(true)` if the connection has failed, `Some(false)` if the connection has not failed,
     /// or `None` if the platform does not support detecting this condition.
     #[inline]
+    #[deprecated]
```
notgull, line 394: Please add a deprecation message here.

notgull (changes requested):

src/lib.rs
irvingoujAtDevolution: updated

examples/tcp_client.rs
notgull: Don't modify the Example too much; just make it so the `is_err`-specific code is guarded.

irvingoujAtDevolution:
please have a look again, I am not exactly sure if this is how you want it.

src/lib.rs
```diff
@@ -391,10 +391,66 @@ impl Event {
...
     pub fn is_connect_failed(&self) -> Option<bool> {
         self.extra.is_connect_failed()
     }
 
+    /// Tells if this event is the result of a connection failure.
+    ///
+    /// This function checks if an error exist,particularlly useful in detecting if TCP connection failed. It corresponds to the `EPOLLERR` event in Linux
+    /// and `CONNECT_FAILED` event in Windows IOCP.
+    ///
+    /// note: in epoll, a tcp connection failure is indicated by EPOLLERR + EPOLLHUP, though just EPOLLERR is enough to indicate a connection failure.
```
notgull, line 407:
```suggestion
    /// ## Caveats
    ///
    /// In `epoll`, a TCP connection failure is indicated by `EPOLLERR` + `EPOLLHUP`, though just `EPOLLERR` is enough to indicate a connection failure.
```

src/lib.rs
```diff
@@ -391,10 +391,66 @@ impl Event {
...
+    ///           None => {
+    ///               println!("not supported in this platform");
+    ///               return Ok(());
+    ///           },
+    ///       };
+    ///       assert!(is_err);
+    ///       println!("bad socket is now in error state");
+    ///   
+    ///       Ok(())
+    ///   }
```
notgull, line 445: Why is this block indented?

src/lib.rs
```diff
@@ -391,10 +391,66 @@ impl Event {
...
+    /// ```  
+    /// Returns `Some(true)` if the connection has failed, `Some(false)` if there is an error,
```
notgull, line 447:
```suggestion
    /// ```  
    ///
    /// Returns `Some(true)` if the connection has failed, `Some(false)` if there is an error,
```

notgull (changes requested):
Overall LGTM, just some minor considerations regarding docs.

src/lib.rs
irvingoujAtDevolution: updated, removed it

src/lib.rs
```diff
@@ -391,10 +391,31 @@ impl Event {
     /// Returns `Some(true)` if the connection has failed, `Some(false)` if the connection has not failed,
     /// or `None` if the platform does not support detecting this condition.
     #[inline]
+    #[deprecated(
+        since = "3.4.0",
+        note = "use `is_err` in combination of is_hub instead, see documentation for `is_err`"
```
notgull, line 396:
```suggestion
        note = "use `is_err` in combination of is_hup instead, see documentation for `is_err`"
```

notgull (commented):
LGTM aside from one minor nitpick

irvingoujAtDevolution:
😢 , @notgull could you please help me look at the wine CI thing? I have no idea how that works

notgull:
Looks like a [spurious error](https://github.com/smol-rs/polling/actions/runs/8363127534/job/22895359605).

notgull (approved):
Thanks!

Issue #191, opened 2024-02-13
unsafe add: will it go back to safe?


keepsimple1:
https://github.com/smol-rs/polling/blob/77b4ed115674f9e7a6f8636cf5e9487a9cb06687/src/lib.rs#L510

Although this seems to have happened almost 6 months ago, it's sad that this API changed to be "unsafe". One of my project relies on this API heavily and we cannot upgrade to the latest "polling" due to this unsafe API. 

Is there any plan to make this safe again? ("polling" is a great library, I love using it).

zeenix:
`unsafe` only means it's **potentially** unsafe and shouldn't stop you from using the library in any way. We can lie that it's safe by not declaring it so but all that will do is give you a false sense of safety.

zeenix:
> `unsafe` only means it's **potentially** unsafe and shouldn't stop you from using the library in any way. We can lie that it's safe by not declaring it so but all that will do is give you a false sense of safety.

Having said that, looking at the [PR that changed it](https://github.com/smol-rs/polling/pull/123), I'm not sure why this method wasn't just changed to take `BorrowedFd` instead. @notgull ?

notgull:
Since the poller uses the reference to the source without considering its lifetime it's essentially taking a raw pointer. I updated the signature to make this clear.

zeenix:
> Since the poller uses the reference to the source without considering its lifetime it's essentially taking a raw pointer. I updated the signature to make this clear.

Thanks for explaining, @notgull. I'm closing this then.

keepsimple1:
That's really fast to close an issue. It will be good to give the user a chance to follow up before closing ;-)

https://github.com/smol-rs/polling/blob/77b4ed115674f9e7a6f8636cf5e9487a9cb06687/src/lib.rs#L615 

I'm wondering, why is `modify()` able to take `AsSource`, not `AsRawSource`?

notgull:
It is able to take a borrowed pointer because it's not kept around.

PR #192, opened 2024-02-15
DRAFT: trying to make add safe

keepsimple1:
To follow up on the issue #191, I'm experimenting with some changes to make `Poller::add` safe.  

Since I'm new to I/O Safety, I probably missed something or maybe totally wrong. But I wanted to post this PR to get comments and find out what I've missed at least. 

This draft change only applied on macOS, i.e. `kqueue` as I'm trying to see if I was on the right direction. Thanks in advance.

### DIFF
--- a/src/kqueue.rs
+++ b/src/kqueue.rs
@@ -84,11 +84,17 @@ impl Poller {
     /// # Safety
     ///
     /// The file descriptor must be valid and it must last until it is deleted.
-    pub unsafe fn add(&self, fd: RawFd, ev: Event, mode: PollMode) -> io::Result<()> {
-        self.add_source(SourceId::Fd(fd))?;
+    pub fn add(&self, fd: BorrowedFd<'_>, ev: Event, mode: PollMode) -> io::Result<()> {
+        let rawfd = fd.as_raw_fd();
+
+        // SAFETY: `rawfd` is valid as it is from `BorrowedFd`. And
+        // this block never closes / deletes `rawfd`.
+        unsafe {
+            self.add_source(SourceId::Fd(rawfd))?;
+        }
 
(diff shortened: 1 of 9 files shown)


src/kqueue.rs
```diff
@@ -84,11 +84,17 @@ impl Poller {
     /// # Safety
     ///
     /// The file descriptor must be valid and it must last until it is deleted.
-    pub unsafe fn add(&self, fd: RawFd, ev: Event, mode: PollMode) -> io::Result<()> {
-        self.add_source(SourceId::Fd(fd))?;
+    pub fn add(&self, fd: BorrowedFd<'_>, ev: Event, mode: PollMode) -> io::Result<()> {
+        let rawfd = fd.as_raw_fd();
+
+        // SAFETY: `rawfd` is valid as it is from `BorrowedFd`. And
+        // this block never closes / deletes `rawfd`.
```
keepsimple1, line 91: Does this comment make sense at all? I'm new to I/O safety, please advise.

notgull:
I don't think it's possible to make `add` safe in the general case; the Windows and `poll()`-based backends can't be made safe. However it is possible to make it safe for certain backends.

keepsimple1:
> I don't think it's possible to make `add` safe in the general case; the Windows 

What if we change `RawSocket` to `BorrowedSocket` in the following? Will that help?
https://github.com/smol-rs/polling/blob/0b4afcaf0a61ee06087d77c490abf8802f91aeff/src/iocp/mod.rs#L174

> and `poll()`-based backends can't be made safe. However it is possible to make it safe for certain backends.

Is it this one? (seems it is still / already safe?)
https://github.com/smol-rs/polling/blob/0b4afcaf0a61ee06087d77c490abf8802f91aeff/src/poll.rs#L102

PR #193, opened 2024-02-17
v3.5.0

notgull:
- Use the `epoll` backend when RedoxOS is enabled. (#190)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Version 3.5.0
+
+- Use the `epoll` backend when RedoxOS is enabled. (#190)
+
 # Version 3.4.0
 
 - Add the ability to identify whether socket connection has failed. (#185)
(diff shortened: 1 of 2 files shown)


taiki-e:
I think it would be safer to raise the minor version since this could potentially cause a subtle change in behavior. (I have in mind the previous regression of async-proces due to a change in implementation on Linux introduced in the patch version.)
Otherwise LGTM.

notgull:
> I think it would be safer to raise the minor version since this could potentially cause a subtle change in behavior. (I have in mind the previous regression of async-proces due to a change in implementation on Linux introduced in the patch version.) Otherwise LGTM.

I doubt that any significant infrastructure has been built on `polling` in Redox, as no one noticed it was using the inefficient `polling` backend until now. Still, it's probably safer.

taiki-e (approved):

PR #194, opened 2024-02-25
Add support for HermitOS

### LINKED ISSUE #177: HermitOS Support
`x86_64-unknown-hermit` appears to have a [`poll`](https://docs.rs/hermit-abi/latest/hermit_abi/fn.poll.html) function similar to Unix. Since we already have a `poll`-based backend we should be able to integrate it into here.

However from a brief look I can't see whether or not it has an equivalent to `pipe` or `eventfd`, which would be needed to implement support here.

notgull:
HermitOS is a microkernel target aiming to provide a simple OS for
virtualized applications. It recently added support for the poll() and
eventfd() system calls, which means we can target it with our poll()
based backend.

Hermit does not have a traditional libc; instead it uses hermit-abi.
However rustix does not support using hermit-abi as its underlying
layer yet. So we have to build a shim layer until it does.

Closes #177
cc bytecodealliance/rustix#1012
cc @stlankes

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -474,7 +474,7 @@ mod syscall {
     pub(super) fn eventfd(count: u64, _flags: EventfdFlags) -> io::Result<OwnedFd> {
         let fd = unsafe { hermit_abi::eventfd(count, 0) };
 
-        if fd == -1 {
+        if fd < 0 {
             Err(io::Error::from_raw_os_error(unsafe {
                 hermit_abi::get_errno()
             }))
@@ -612,7 +612,7 @@ mod syscall {
     /// Convert a number to an actual result.
     #[inline]
     fn cvt(len: isize) -> io::Result<usize> {
-        if len == -1 {
+        if len < 0 {
             Err(io::Error::from_raw_os_error(unsafe {
                 hermit_abi::get_errno()
             }))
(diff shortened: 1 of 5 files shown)


stlankes:
Looks good from my side! However, we have an issue `last_os_error()` and it doesn't work like expected. But we are able to fix it. I  will give you a hit, it it works like expected.

notgull:
> Looks good from my side! However, we have an issue `last_os_error()` and it doesn't work like expected. But we are able to fix it. I will give you a hit, it it works like expected.

Could you elaborate? What error is being returned?

stlankes:
In `std` we determine the last error code in a different way and ignored `last_os_error`. We just don't use it. To solve it, we submit rust-lang/rust#121765. In the current stage, I would directly use [hermit-abi](https://github.com/hermit-os/hermit-rs/blob/main/hermit-abi/src/lib.rs#L342) to get the last error code.

notgull:
@stlankes I've fixed it to use `get_errno` instead of `last_os_error`, can you take another look?

src/poll.rs
```diff
@@ -475,7 +475,9 @@ mod syscall {
         let fd = unsafe { hermit_abi::eventfd(count, 0) };
 
         if fd == -1 {
```
stlankes, line 477: ` if fd == -1 {`

You should use ` if fd < 0 {`. In our case, the return values could be another negative number.

src/poll.rs
```diff
@@ -611,7 +613,9 @@ mod syscall {
     #[inline]
     fn cvt(len: isize) -> io::Result<usize> {
         if len == -1 {
```
stlankes, line 615: See comments above. `if len < 0 {` is the correct way. All other parts are in my opinion correct.

stlankes (changes requested):

PR #196, opened 2024-03-23
v3.6.0

notgull:
- Add an `is_err` method to `Event` to tell when an error has occurred. (#189)
- Deprecate the `is_connect_failed` function. (#189)
- Add support for HermitOS to `polling`. (#194)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Version 3.6.0
+
+- Add an `is_err` method to `Event` to tell when an error has occurred. (#189)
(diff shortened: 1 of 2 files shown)


taiki-e (approved):

taiki-e:
It may also be worth mentioning the deprecation of is_connect_failed.

notgull:
> It may also be worth mentioning the deprecation of is_connect_failed.

Good point, I've added it.

taiki-e (approved):

PR #197, opened 2024-04-17
feat: ported to Vita target

### LINKED ISSUE #160: Add support for PS Vita
Related: https://github.com/tokio-rs/mio/pull/1721

This will probably need to be tacked onto the `poll()` backend like ESP-IDF is.

nikarh:
Fixes #160

### Notes
- `signal-hook` does not compile and won't work on Vita, `pthread_kill` is a stub in Vita `newlib`
- Pipes in Vita newlib do not guarantee that after the write completes, the payload is immediately available on the other side of the pipe. This PR temporarily removes the `NONBLOCK` flag from the read side of the pipe when `wait` detects `notified` before polling

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -222,7 +222,28 @@ impl Poller {
                 if self.notified.swap(false, Ordering::SeqCst) {
                     // `notify` will have sent a notification in case we were polling. We weren't,
                     // so remove it.
-                    return self.notify.pop_notification();
+
+                    // Pipes on Vita do not guarantee that after `write` call succeeds, the
+                    // data becomes immediately available for reading on the other side of the pipe.
+                    // To ensure that the notification is not lost, the read side of the pipe is temporarily
+                    // switched to blocking for a single `read` call.
+                    #[cfg(target_os = "vita")]
+                    rustix::fs::fcntl_setfl(
+                        &self.notify.read_pipe,
+                        rustix::fs::fcntl_getfl(&self.notify.read_pipe)?
+                            & !rustix::fs::OFlags::NONBLOCK,
+                    )?;
+
+                    let notification = self.notify.pop_notification();
+
+                    #[cfg(target_os = "vita")]
+                    rustix::fs::fcntl_setfl(
+                        &self.notify.read_pipe,
+                        rustix::fs::fcntl_getfl(&self.notify.read_pipe)?
+                            | rustix::fs::OFlags::NONBLOCK,
+                    )?;
+
+                    return notification;
                 } else if self.waiting_operations.load(Ordering::SeqCst) == 0 {
                     break;
                 }
@@ -646,7 +667,7 @@ mod notify {
     pub(super) struct Notify {
         /// The file descriptor of the read half of the notify pipe. This is also stored as the first
         /// file descriptor in `fds.poll_fds`.
-        read_pipe: OwnedFd,
+        pub(super) read_pipe: OwnedFd,
         /// The file descriptor of the write half of the notify pipe.
         ///
         /// Data is written to this to wake up the current instance of `Poller::wait`, which can occur when the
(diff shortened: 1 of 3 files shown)


src/poll.rs
```diff
@@ -222,7 +222,28 @@ impl Poller {
...
+                    let notification = self.notify.pop_notification();
+
+                    #[cfg(target_os = "vita")]
+                    rustix::fs::fcntl_setfl(
+                        &self.notify.read_pipe,
+                        rustix::fs::fcntl_getfl(&self.notify.read_pipe)?
+                            | rustix::fs::OFlags::NONBLOCK,
+                    )?;
+
+                    return notification;
```
notgull, line 246: I think this could be inlined into `pop_notification`.

notgull (changes requested):
Can we add a test for this in CI? Just a `cargo check` would suffice.

Also, do you have a way of testing this on the actual Vita software?

nikarh:
> Can we add a test for this in CI? Just a `cargo check` would suffice.
> 
> Also, do you have a way of testing this on the actual Vita software?

Running tests requires an actual piece of hardware (no way to run in qemu, since Vita runs a proprietary OS that demands specific hardware). Cargo check though is absolutely doable, will add it.

Currently, I run tests on a physical device using `cargo-vita` tool which does these things in a single command:
- Compiles rust test code and links it with VITASDK toolchain gcc to an ELF (this is essentially cargo build with a bunch of flags against the target, linker is defined in the [target spec](https://github.com/rust-lang/rust/blob/13e63f7490109eecf905fa51190eacbe8051aefd/compiler/rustc_target/src/spec/targets/armv7_sony_vita_newlibeabihf.rs#L37))
- Transforms the usual ELF into Sony ELF (by calling a number of tools from VITASDK toolchain)
- Uploads the Sony ELF to Vita via FTP (usually all devs who target Vita run a userspace module on the device that keeps an FTP server and a TCP command server)
- Sends a "run this app" command to the TCP command server

At the same time, I run another plain text TCP socket server on my computer, and Vita runs a module that pipes stdin and stdout to that TCP socket, making it possible to see test results.

src/poll.rs
nikarh: Done

nikarh:
Fixed the comment, added check to CI and enabled CI check for tier 3 nightly targets

src/poll.rs
```diff
@@ -699,8 +699,24 @@ mod notify {
...
+            // data becomes immediately available for reading on the other side of the pipe.
+            // To ensure that the notification is not lost, the read side of the pipe is temporarily
+            // switched to blocking for a single `read` call.
+            #[cfg(target_os = "vita")]
+            rustix::fs::fcntl_setfl(
+                &self.read_pipe,
+                rustix::fs::fcntl_getfl(&self.read_pipe)? & !rustix::fs::OFlags::NONBLOCK,
+            )?;
+
             read(&self.read_pipe, &mut [0; 1])?;
```
notgull, line 712: I'm worried that, if `read` errors out, the system call setting the pipe back to non-blocking won't be called. Instead of the `?` here, grab the result to make sure that the non-blocking syscall is still called.

notgull (changes requested):
I'm assuming the tests pass when you run it on actual hardware, right?

nikarh:
> I'm assuming the tests pass when you run it on actual hardware, right?

Yes, I run them on the actual hardware

src/poll.rs
nikarh: Makes sense, done

notgull (approved):
Thanks! NetBSD failure looks spurious

.github/workflows/ci.yml
```diff
@@ -81,15 +81,16 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
+        rust: [nightly, stable]
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust
-      run: rustup update stable
+      run: rustup update ${{ matrix.rust }}
```
taiki-e, line 88: Instead of adding `+nightly` to all cargo command calls, I recommended using ` && rustup default ${{ matrix.rust }}` here.

PR #205, opened 2024-05-29
Fix new nightly errors

notgull:
- Use i32::MAX instead of std::i32::MAX
- Add "expected cfg" lints to Cargo.toml

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,9 @@ exclude = ["/.*"]
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(docsrs)', 'cfg(polling_test_poll_backend)', 'cfg(polling_test_epoll_pipe)'] }
+
 [dependencies]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -17,6 +17,9 @@ exclude = ["/.*"]
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(docsrs)', 'cfg(polling_test_poll_backend)', 'cfg(polling_test_epoll_pipe)'] }
```
taiki-e, line 21:
```suggestion
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(polling_test_poll_backend)', 'cfg(polling_test_epoll_pipe)'] }
```

cfg(docsrs) is an official cfg set by docs.rs. https://github.com/rust-lang/cargo/blob/431db31d0dbeda320caf8ef8535ea48eb3093407/src/cargo/core/compiler/mod.rs#L1351

taiki-e:
LGTM aside from the above nit.

taiki-e (approved):

PR #217, opened 2024-09-14
Add support for visionOS

madsmtm:
[In the future](https://github.com/rust-lang/rust/issues/100343#issuecomment-2065159934) you'll probably be able to use `target_family = "darwin"` instead, but for now `target_vendor = "apple"` should suffice.


taiki-e (approved):
Thanks!

When I first looked at "Deprecate target_vendor" MCP I was under the impression that it might be done in such a way as to break existing Apple-related code, so I was preferring to avoid `target_vendor = "apple"`, but from the latest discussion it seems okay.

taiki-e:
Could you update docs/comments too?

https://github.com/smol-rs/polling/blob/7531821ae2d61f967bb66aab9eb18310a8831a3d/src/lib.rs#L5
https://github.com/smol-rs/polling/blob/7531821ae2d61f967bb66aab9eb18310a8831a3d/src/kqueue.rs#L1
https://github.com/smol-rs/polling/blob/7531821ae2d61f967bb66aab9eb18310a8831a3d/README.md?plain=1#L16

taiki-e (approved):

PR #221, opened 2024-12-04
On kqueue platforms, make it so the EV_EOF flag corresponds to the "is_interrupt" flag.

### LINKED ISSUE #213: Make it so `Event::is_interrupt` corresponds to EV_EOF on kqueue platforms
Apparently this is pretty close to an interruption

notgull:
Closes #213

### DIFF
--- a/src/kqueue.rs
+++ b/src/kqueue.rs
@@ -110,16 +110,21 @@ impl Poller {
 
         let mode_flags = mode_to_flags(mode);
 
-        let read_flags = if ev.readable {
+        let mut read_flags = if ev.readable {
             kqueue::EventFlags::ADD | mode_flags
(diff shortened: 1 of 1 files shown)


fogti:
Can you point to any documentation / man-page or mailing list post which proves this? I couldn't find anything indicating this association in `kqueue(2)` of FreeBSD 14.2. (It is of course still a good idea to be able to specify the flag, I just don't see why this is termed "interrupt"; on the other hand, relation to HUP is a bit stronger, because it gets emitted when the end of the stream is reached, e.g. because all writers of a FIFO or such closed)

notgull:
I swear that I saw some code using it like this, but I can't seem to find it now. Standby.

PR #226, opened 2025-02-18
add `wait_deadline` function

joshyrobot:
A handy function if your code mostly relies on deadlines rather than timeouts.

Also changes the internal pollers to use deadlines and has them calculate timeouts just before the relevant syscall. This prevents a couple extra calls to `Instant::now()`, hypothetically makes the timeout more accurate, and in my opinion is a little bit easier to reason about.

(I only compiled and tested epoll so here's hoping CI goes well)


src/lib.rs
```diff
@@ -729,19 +729,28 @@ impl Poller {
...
+    }
+
+    /// Waits for at least one I/O event and returns the number of new events, with an optional
+    /// deadline.
+    ///
+    /// See [`wait()`][`Poller::wait()`] for more details.
+    pub fn wait_deadline(
+        &self,
+        events: &mut Events,
+        deadline: Option<Instant>,
```
taiki-e, line 745: I think allowing `wait_deadline(e, None)` is a bit odd. We can always take `Instant` instead of `Option<Instant>`.

joshyrobot: Does the new approach of `pub fn wait_deadline(... deadline: Instant)` and `fn wait_impl(... deadline: Option<Instant>)` look agreeable? At some point it needs to be possible to pass `None` otherwise `wait` would need to duplicate the implementation.

PR #230, opened 2025-03-18
Port to rustix 1.0.

sunfishcode:
rustix 1.0 introduced a new [`Buffer`] API which is used in `epoll::wait`, `kqueue`, and `polling`, so this patch updates the code to use the new [`spare_capacity`] mechanism for writing to the spare capacity of a `Vec`.

rustix 1.0 also changes timeout parameters from `i32` milliseconds to `Timespec` seconds plus nanoseconds, which is more general and aligns with newer Linux syscalls such as [`epoll_pwait2`].

This also fixes a bug in the `wait` function; in rustix 0.38, calling `epoll::wait` would clear the events list before adding new events, however polling's documentation [states] that "New events will be appended to `events`." Using `spare_capacity` in rustix 1.0, the list is not cleared first, so events are appended.

Closes #234.

[`Buffer`]: https://docs.rs/rustix/latest/rustix/buffer/trait.Buffer.html
[`spare_capacity`]: https://docs.rs/rustix/latest/rustix/buffer/fn.spare_capacity.html
[`epoll_pwait2`]: https://www.man7.org/linux/man-pages/man2/epoll_pwait2.2.html
[states]: https://docs.rs/polling/latest/polling/struct.Poller.html#method.wait

### DIFF
--- a/src/epoll.rs
+++ b/src/epoll.rs
@@ -9,13 +9,13 @@ use rustix::event::{eventfd, EventfdFlags};
 #[cfg(not(target_os = "redox"))]
 use rustix::time::{
     timerfd_create, timerfd_settime, Itimerspec, TimerfdClockId, TimerfdFlags, TimerfdTimerFlags,
-    Timespec,
 };
 
-use rustix::event::epoll;
+use rustix::buffer::spare_capacity;
+use rustix::event::{epoll, Timespec};
 use rustix::fd::OwnedFd;
 use rustix::fs::{fcntl_getfl, fcntl_setfl, OFlags};
-use rustix::io::{fcntl_getfd, fcntl_setfd, read, write, FdFlags};
+use rustix::io::{fcntl_getfd, fcntl_setfd, read, write, Errno, FdFlags};
 use rustix::pipe::{pipe, pipe_with, PipeFlags};
 
 use crate::{Event, PollMode};
@@ -196,22 +196,19 @@ impl Poller {
         #[cfg(target_os = "redox")]
         let timer_fd: Option<core::convert::Infallible> = None;
 
-        // Timeout in milliseconds for epoll.
-        let timeout_ms = match (timer_fd, timeout) {
-            (_, Some(t)) if t == Duration::from_secs(0) => 0,
-            (None, Some(t)) => {
-                // Round up to a whole millisecond.
-                let mut ms = t.as_millis().try_into().unwrap_or(i32::MAX);
-                if Duration::from_millis(ms as u64) < t {
-                    ms = ms.saturating_add(1);
-                }
-                ms
-            }
-            _ => -1,
+        // Timeout for epoll.
+        let timeout = match (timer_fd, timeout) {
+            (_, Some(t)) if t == Duration::from_secs(0) => Some(Timespec::default()),
+            (None, Some(t)) => Some(Timespec::try_from(t).map_err(|_| Errno::INVAL)?),
+            _ => None,
         };
 
         // Wait for I/O events.
-        epoll::wait(&self.epoll_fd, &mut events.list, timeout_ms)?;
+        epoll::wait(
+            &self.epoll_fd,
+            spare_capacity(&mut events.list),
+            timeout.as_ref(),
+        )?;
         tracing::trace!(
             epoll_fd = ?self.epoll_fd.as_raw_fd(),
             res = ?events.list.len(),
@@ -306,7 +303,7 @@ fn write_flags() -> epoll::EventFlags {
 
 /// A list of reported I/O events.
 pub struct Events {
-    list: epoll::EventVec,
+    list: Vec<epoll::Event>,
 }
 
 unsafe impl Send for Events {}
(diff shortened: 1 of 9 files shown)


sunfishcode:
I've now also tested this on illumos, to test the `port` backend. Everything passes except for the `insert_twice` test, however this doesn't appear to be a regression. The code is checking that inserting a fd twice gets an `AlreadyExists` error, however  `port_associate` doesn't fail is the fd is already associated, and the port.rs code doesn't have any extra bookkeeping.

src/epoll.rs
```diff
@@ -196,22 +196,19 @@ impl Poller {
...
-                if Duration::from_millis(ms as u64) < t {
-                    ms = ms.saturating_add(1);
-                }
-                ms
-            }
-            _ => -1,
+        // Timeout for epoll.
+        let timeout = match (timer_fd, timeout) {
+            (_, Some(t)) if t == Duration::from_secs(0) => Some(Timespec::default()),
+            (None, Some(t)) => Some(Timespec::try_from(t).map_err(|_| Errno::INVAL)?),
```
taiki-e, line 202: This will return an error in cases that did not previously return an error, right?

src/poll.rs
```diff
@@ -454,14 +441,29 @@ fn cvt_mode_as_remove(mode: PollMode) -> io::Result<bool> {
...
     #[cfg(target_os = "espidf")]
     pub(super) use rustix::io::{read, write};
+    use std::io;
+    use std::time::Duration;
+
+    /// Safe wrapper around the `poll` system call.
+    pub(super) fn poll(fds: &mut [PollFd<'_>], timeout: Option<Duration>) -> io::Result<usize> {
+        // Timeout for `poll`.
+        let timeout = match timeout {
+            Some(timeout) => Some(Timespec::try_from(timeout).map_err(|_| Errno::INVAL)?),
```
taiki-e, line 461: Same here.

src/poll.rs
```diff
@@ -503,12 +506,25 @@ mod syscall {
...
+    pub(super) fn poll(fds: &mut [PollFd<'_>], timeout: Option<Duration>) -> io::Result<usize> {
+        let timeout_ms = match timeout {
+            Some(timeout) => {
+                // Round up to a whole millisecond.
+                let mut ms = timeout.as_millis().try_into().unwrap_or(std::u64::MAX);
+                if Duration::from_millis(ms) < timeout {
+                    ms = ms.saturating_add(1);
+                }
+                ms.try_into()
+                    .map_err(|_| io::Error::from(io::ErrorKind::InvalidInput))?
```
taiki-e, line 518: Same here.

src/port.rs
```diff
@@ -112,8 +113,19 @@ impl Poller {
         );
         let _enter = span.enter();
 
+        // Timeout for `port::getn`.
+        let timeout = match timeout {
+            Some(t) => Some(Timespec::try_from(t).map_err(|_| Errno::INVAL)?),
```
taiki-e, line 118: Same here.

taiki-e (changes requested):
Thanks! This is generally LGTM.

src/epoll.rs
fogti: This will produce an error if the seconds part of "t" doesn't fit into `i64`.

fogti:
For the mentioned problem, I think it would be a better idea to put `rustix::event::Timespec` into our public API, and re-export it from the crate root, instead of wrapping the fallible conversion.

taiki-e:
I think the generally preferred behavior regarding time overflow is saturation or close to it.
https://github.com/smol-rs/async-io/pull/87
https://github.com/smol-rs/async-io/issues/86#issuecomment-1221090184

src/epoll.rs
fogti: Note how big `2^63 - 1` (the largest number representable using `i64`) is: https://www.wolframalpha.com/input?i=2%5E63+-+1+seconds

It is not useful to support larger values, at least not for timeouts.

taiki-e: It doesn't matter how “unlikely it is to be reached by natural increase”; people rather use such values because of their nature.

See the issues linked from the issues I linked to, e.g. https://github.com/crossbeam-rs/crossbeam/issues/795.

sunfishcode:
Makes sense; I've now changed the code to handle `Duration`->`Timespec` overflow by saturating to an infinite timeout.

Looking at the log, the i686-unknown-freebsd failure looks spurious.

src/poll.rs
taiki-e: @sunfishcode This does not seem to have been addressed yet. Or am I missing something?

sunfishcode: Ah, you're right, I had missed that one. It's now addressed.

sunfishcode:
The i686-unknown-freebsd failure is real; I'm working on a fix.

taiki-e (approved):
Thanks!

fogti:
Sorry, I currently don't have time to properly review this (as I'm trying to finish a thesis); maybe after next week. I wish I'd knew how to test this for Hermit, but I currently don't.

fogti (approved):
Manual test suite run passes on `x64_64-unknown-freebsd` and `x86_64-unknown-linux-gnu`.

Please write a `CHANGELOG.md` entry for this, given that it changes how event vectors are handled: they aren't cleared anymore, this is a bugfix that brings the code back in line with what it should be, but that itself doesn't prevent downstream from accidentally ending up relying on the former, buggy behavior.

taiki-e:
Published in 3.8.0.

PR #235, opened 2025-04-10
Implement AsRawFd for Poller on Redox OS

jackpot51:
Redox OS is one of the epoll platforms so I added it next to the other epoll platforms.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -963,6 +963,7 @@ impl fmt::Debug for Events {
     any(
         target_os = "linux",
         target_os = "android",
+        target_os = "redox",
         target_os = "illumos",
         target_os = "solaris",
         target_vendor = "apple",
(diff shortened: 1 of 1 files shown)


zeenix:
> I wish we had some tests for Redox. Maybe some day Github will support Redox runners. :)

Hmm.. actually seems we [already check the build for niche OSes like netbsd](https://github.com/smol-rs/polling/blob/master/.github/workflows/ci.yml#L164). Perhaps good idea to do that for redox too? :thinking:

jackpot51:
It should be possible, either with `cargo check` or with `redoxer`.

zeenix:
> It should be possible, either with `cargo check` or with `redoxer`.

`redoxer` seems nice. If you could add the CI, that would be great.

jackpot51:
Unfortunately we are in the midst of completing some large projects that can make redoxer unreliable.

jackpot51:
Can a new release be made with this change?

PR #240, opened 2025-07-05
Update to socket2 v0.6

Thomasdezeeuw:
Closes #241


taiki-e (approved):
Thanks for the PR!

socket2 is a dec-dependency so it does not affect MSRV, it is tracing that is breaking the MSRV build and we plan to fix this by making the dependency on tracing optional.

Issue #255, opened 2025-12-20
Make the "add" function safe


notgull:
One of my regrets from when I was more active is that I left the "add" function unsafe. I feel that there should be a way to make this function safe without any safety issues.

The main problem platform is `kqueue`. `epoll` automatically deregisters the FD if it's closed,  so no issues there. Windows uses generational handles for its sockets, so there's no chance of a socket being reused. But `kqueue` and the `poll` backend still poll the file descriptor even after it's closed. This causes a use-after-free.

One thought I have is that this is basically not an issue. In most cases, this leads to a fatal I/O error. At worst it can lead to spurious events being returned.

That being said, there's also something to be said that this is still fundamentally using a resource after it's been deallocated. In addition if Miri ever got the capacity for checking file descriptors, it would raise a red flag here.

We could have a safe version of the `add` function where it can only add file descriptors known to outlive the `Poller`. However, in 99% of cases the `Poller` is `'static`, so this would have limited usefulness.

So, can we make this `add` function safe? Let's discuss. I'm leaning towards yes.

PR #256, opened 2025-12-23
Make add() safe

notgull:
As per discussion in #255, the add() function can be made safe with no
other modifications to the API. This allows the `polling` crate to be
used entirely safely again.

I don't like `AsRawSource` still being in the public API, but that
can't really be helped at this point. If we do another breaking change
in the future this can be fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,15 +136,19 @@ jobs:
       run: |
         rustup target add x86_64-unknown-redox
         cargo check --target x86_64-unknown-redox
-    - name: HermitOS
-      if: startsWith(matrix.rust, 'nightly') && startsWith(matrix.os, 'ubuntu')
-      run: cargo check -Z build-std --target x86_64-unknown-hermit
+    # Bring this back once Hermit stdlib bugs are fixed
+    # See: https://github.com/rust-lang/rust/issues/150294
+    #- name: HermitOS
+    #  if: startsWith(matrix.rust, 'nightly') && startsWith(matrix.os, 'ubuntu')
+    #  run: cargo check -Z build-std --target x86_64-unknown-hermit
     - name: Check haiku
       if: startsWith(matrix.rust, 'nightly') && startsWith(matrix.os, 'ubuntu')
       run: cargo check -Z build-std --target x86_64-unknown-haiku
-    - name: Check vita
-      if: startsWith(matrix.rust, 'nightly') && startsWith(matrix.os, 'ubuntu')
-      run: cargo check -Z build-std --target armv7-sony-vita-newlibeabihf
+    # Bring this back once Vita stdlib bugs are fixed
+    # See: https://github.com/rust-lang/rust/pull/150297
+    #- name: Check vita
+    #  if: startsWith(matrix.rust, 'nightly') && startsWith(matrix.os, 'ubuntu')
+    #  run: cargo check -Z build-std --target armv7-sony-vita-newlibeabihf
     - name: Check ESP-IDF
       if: startsWith(matrix.rust, 'nightly') && startsWith(matrix.os, 'ubuntu')
       run: cargo check -Z build-std --target riscv32imc-esp-espidf


notgull:
Is this an stdlib bug? Filed https://github.com/rust-lang/rust/issues/150294

zeenix (changes requested):
The change itself seems good. Please also add details to commit messages so all details of changes are available offline in git history. 🙏

src/iocp/mod.rs
```diff
@@ -169,12 +169,7 @@ impl Poller {
     /// # Safety
     ///
     /// The socket must be a valid socket and must last until it is deleted.
-    pub(super) unsafe fn add(
-        &self,
-        socket: RawSocket,
-        interest: Event,
-        mode: PollMode,
-    ) -> io::Result<()> {
+    pub(super) fn add(&self, socket: RawSocket, interest: Event, mode: PollMode) -> io::Result<()> {
```
taiki-e: Same here.

src/kqueue.rs
```diff
@@ -86,11 +86,11 @@ impl Poller {
     /// # Safety
     ///
     /// The file descriptor must be valid and it must last until it is deleted.
-    pub unsafe fn add(&self, fd: RawFd, ev: Event, mode: PollMode) -> io::Result<()> {
+    pub fn add(&self, fd: RawFd, ev: Event, mode: PollMode) -> io::Result<()> {
```
taiki-e: Same here.

src/os/iocp.rs
```diff
@@ -98,7 +98,7 @@ pub trait PollerIocpExt: PollerSealed {
     /// assert_eq!(events.len(), 1);
     /// assert_eq!(events.iter().next().unwrap(), Event::all(0));
     /// ```
-    unsafe fn add_waitable(
+    fn add_waitable(
```
taiki-e: Same here.

taiki-e (changes requested):

src/kqueue.rs
```diff
@@ -86,11 +86,11 @@ impl Poller {
...
         self.add_source(SourceId::Fd(fd))?;
 
         // File descriptors don't need to be added explicitly, so just modify the interest.
-        self.modify(BorrowedFd::borrow_raw(fd), ev, mode)
+        self.modify(unsafe { BorrowedFd::borrow_raw(fd) }, ev, mode)
```
taiki-e: Since the responsibility for ensuring safety has shifted from the caller to us, I feel we need to leave a comment here explaining that this is safe. (Something like `SAFETY: See https://github.com/smol-rs/polling/issues/255.` is probably fine, though.)

src/lib.rs
```diff
@@ -520,13 +509,12 @@ impl Poller {
...
-    /// unsafe {
-    ///     poller.add(&source, Event::all(key))?;
-    /// }
+    /// poller.add(&source, Event::all(key))?;
     /// poller.delete(&source)?;
     /// # std::io::Result::Ok(())
     /// ```
-    pub unsafe fn add(&self, source: impl AsRawSource, interest: Event) -> io::Result<()> {
+    // TODO: At the next breaking change, change `AsRawSource` to `AsSource`
+    pub fn add(&self, source: impl AsRawSource, interest: Event) -> io::Result<()> {
```
hanna-kruppe: I'm not familiar with the `polling` crate, but I don't think this is signature is sound w.r.t. I/O safety, even outside the "backend may still poll the fd after it's closed" aspect discussed in #255. Standard library documentation [says explicitly](https://doc.rust-lang.org/std/io/index.html#io-safety):

> To uphold I/O safety, it is crucial that no code acts on file descriptors it does not own or borrow, and no code closes file descriptors it does not own. In other words, a safe function that takes a regular integer, treats it as a file descriptor, and acts on it, is unsound.

... and this API does exactly that, by taking `impl AsRawSource` which includes `RawFd = c_int`. For example, `poller.add(666, ev)` acts on the regular integer 666 as if was an fd.

Note that it's not enough that there happens to be an fd with this numeric value at the time of call. Whoever "owns" the fd must consent to giving up exclusive ownership for as long as the poller may still use it, by either passing ownership of the fd (unlikely since they'd still want to act on it when getting events) or borrowing the fd to the poller. Conversely, they must promise to not close the fd while the poller is using it, otherwise *something else* could reuse the fd for something that's supposed to be owned). For example, a function like this would violate I/O safety because a caller who has an `OwnedFd` would be allowed to close the fd after calling this function:

```rust
fn add_to_poller(fd: BorrowedFd<'_>, poller: &Poller, ev: Event, mode: PollMode) -> io::Result<()> {
    // SAFETY: `fd` is a valid file descriptor
    unsafe {
        poller.add(fd.as_raw_fd(), ev, mode)
    }
}
```

notgull: As discussed in #255, the worst thing that can happen at the `polling` boundary is that the wrong event gets delivered. This is an acceptable compromise to ensure public API safety, especially since (as discussed up-thread) I/O safety is more of a suggestion.

hanna-kruppe: As I said, I'm not very familiar with the internals of `polling` nor with all its backends. If there's really no way that spurious polling of a random integer that happens to match the numeric value of someone's `OwnedFd` can negatively impact what the owner is doing with their fd, fair enough. But this doesn't really seem obvious to me from the discussion in #255 and from what I know about e.g. epoll.

Past discussion [here](https://github.com/bytecodealliance/rustix/pull/487) and [here](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/I.2FO.20safety.20for.20epoll) was about what happens if you have an fd, add it to an epoll instance, and then close the fd. I understand why that is fine. However, it's not clear to me that it's a non-issue for other backends where the OS won't automatically remove the fd on close. Even with epoll, it seems surprising and potentially problematic if an `OwnedFd` you just created and never shared with anyone was already added (causing you to get EEXIST when trying to add it) or modified/deleted (e.g., causing you to miss events you'd otherwise be guaranteed to get). There's also at least one epoll flag (EPOLLEXCLUSIVE) that has spooky action at a distance between different epoll instances that all contain the same fd, though I think this one's harmless as long as `polling` doesn't use it itself (but will that also be true for the next epoll flag Linux adds in the future?).

I don't think it's necessarily unreasonable to say these kinds of problems aren't safety concerns and ergonomics win. But that does means that polling is potentially incompatible with other libraries (existing or yet-to-be-written) that try to do fancy things with polling based on I/O safety. If you say the worst thing that can happen is delivering wrong events, I can see where you're coming from, but it seems a bit too simplistic when considering *composability* of separately-developed safe abstractions.

> especially since (as discussed up-thread) I/O safety is more of a suggestion.

I don't see any discussion on this PR (or in #255 for that matter) that I would summarize as "I/O safety is more of a suggestion". Perhaps you're referring to @RalfJung [explaining](https://github.com/smol-rs/polling/issues/255#issuecomment-3682824058) that I/O safety is library invariants and thus can't be checked by miri, and that there's currently no satisfying way to account for fds passed by env variables. But taking that to mean "I/O safety is just a suggestion" seems like a big leap to me.

hanna-kruppe: I'm not actually sure if edge triggered polling is relevant in this case. At least with the epoll backend, my understanding is that epoll instances are independent w.r.t. event delivery (other than EPOLLEXCLUSIVE) and level/edge choice, i.e., edge-triggered polling of an fd *in your own epoll instance* still gets all the events even if the same fd was added to *another* epoll instance via `polling`. I don't think `polling` itself makes any API guarantees that would make it sound to build any safety invariants on reasoning like "I added it fd that I own and didn't share with anyone to this `Poller`, thus [...]" -- e.g., if a bug in `polling` causes it to sometimes not poll an fd that was registered, this probably wouldn't be treated as a *soundness* bug in the library.

I don't have encyclopedic knowledge of different polling APIs (and even my epoll knowledge is limited) so maybe I'm wrong or there are similar problems with other polling backends. But as long as there's no side effects on *other uses* of the same fd, then it's probably harmless w.r.t. I/O safety. However, that's a rather open-ended question, and the answer may change as operating systems add new features and APIs (cf. EPOLLEXCLUSIVE added in Linux 4.5). Asking libs-api seems like a good idea in any case.

RalfJung: Ah, right, the edge state is per epoll instance. That part should be fine then. But as you say there may be other interactions.

zeenix: @notgull Reading through the arguments here, I'm much less certain now that this is a good idea. Sorry!

